PR #39, opened 2019-11-04
Set reads to unordered for dense and sparse

Shelnutt2:
Set reads to unordered for dense and sparse


Shelnutt2:
unordered isn't currently allowed for dense, closing.

PR #47, opened 2019-11-18
Remove all use of table locks, none are needed

Shelnutt2:
Remove all use of table locks, none are needed

### DIFF
--- a/mytile/ha_mytile.cc
+++ b/mytile/ha_mytile.cc
@@ -54,27 +54,6 @@
 // Handler for mytile engine
 handlerton *mytile_hton;
 
-/**
-  @brief
-  Function we use in the creation of our hash to get key.
-*/
-
-#ifdef HAVE_PSI_INTERFACE
-static PSI_mutex_key ex_key_mutex_mytile_share_mutex;
-
-static PSI_mutex_info all_mytile_mutexes[] = {
-    {&ex_key_mutex_mytile_share_mutex, "mytile_share::mutex", 0}};
-
-static void init_mytile_psi_keys() {
-  const char *category = "mytile";
-  int count;
-
-  count = array_elements(all_mytile_mutexes);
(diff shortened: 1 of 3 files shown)


normanb:
I did the following to test READ locks;

```create table test_array (   dim0 integer DIMENSION=1 lower_bound="0" upper_bound="100" tile_extent="10",   dim1 integer DIMENSION=1 lower_bound="0" upper_bound="100" tile_extent="10",   attr1 varchar(255) ) engine=MyTile array_type='SPARSE';

insert into  test_array (dim0, dim1, attr1) 
values (1, 1, "cell 1"), (1, 2, "cell 2"), (2, 1, "cell 3");

LOCK TABLE test_array READ;
```
And then within that process;

```
LOCK TABLE test_array READ;
insert into  test_array (dim0, dim1, attr1)  values (3, 3, "cell 4");
ERROR 1099 (HY000): Table 'test_array' was locked with a READ lock and can't be updated
```
Which is the expected behaviour.

The from a different process (connected to existing docker container)

```
insert into  test_array (dim0, dim1, attr1)  values (3, 3, "cell 4");
```

And checking the processlist in the original process

```
| 11 | root        | localhost | test | Query   |   19 | Waiting for table level lock | insert into  test_array (dim0, dim1, attr1)  values (3, 3, "cell 4") |    0.000 |
```
Again as expected.

Releasing the lock and the table is updated.

:+1:  from me

Shelnutt2:
@normanb thanks for the testing. Calling lock tables will still effect us as you've seen because this adds a lock at the mariadb level. Removal of the mytile share/locks in the code really just effect the call open table and subsequently the handler will not have a lock data structure, so the handler itself won't attempt any lock verification. So concurrent writes/reads won't cause any contention except if the user manually issues a `LOCK TABLE` command or `FLUSH TABLE WITH READ LOCK`

PR #48, opened 2019-11-20
Add Primary Key Support

Shelnutt2:
This adds primary key support along with index condition pushdown, and three new session variables for toggling support of condition pushdown, estimating records in the array on open and for dimensions as primary keys.

If conditions are not pushdown either with normal condition pushdow or index condition pushdown we fall back to pushdown the key scan conditions.

```
MariaDB [test]> explain select count(*), id FROM test1 GROUP BY id;
+------+-------------+----------------------+-------+---------------+---------+---------+------+---------+-------------+
| id   | select_type | table                | type  | possible_keys | key     | key_len | ref  | rows    | Extra       |
+------+-------------+----------------------+-------+---------------+---------+---------+------+---------+-------------+
|    1 | SIMPLE      | test1 | index | NULL          | PRIMARY | 8       | NULL | 1095000 | Using index |
+------+-------------+----------------------+-------+---------------+---------+---------+------+---------+-------------+
1 row in set (0.002 sec)
MariaDB [test]> explain select count(*), id FROM test1_no_keys GROUP BY id;
+------+-------------+------------------------------+------+---------------+------+---------+------+---------+---------------------------------+
| id   | select_type | table                        | type | possible_keys | key  | key_len | ref  | rows    | Extra                           |
+------+-------------+------------------------------+------+---------------+------+---------+------+---------+---------------------------------+
|    1 | SIMPLE      | test1_no_keys | ALL  | NULL          | NULL | NULL    | NULL | 1095000 | Using temporary; Using filesort |
+------+-------------+------------------------------+------+---------------+------+---------+------+---------+---------------------------------+
1 row in set (0.005 sec)
```

Notice that from the table which does not have primary keys mariadb does a temp table and file sort, where when we tell it the primary key it knows the uniqueness already and avoid a temp table and sort call.

Performance wise on my laptop with primary key indications mariadb does the group by in 2.4 seconds, without primary key it takes 3.5 seconds. This is will 1 million records which is 15MB on disk (so trivially small).

Pushdown now happens in 1 of 3 places. We can intercept the where clause directly (there is now a config option for enable/disable this, useful mostly for testing). We can also intercept "index conditions". Index conditions are like the main predicate conditions EXCEPT they have gone through the optimize cycle and are "limited to the fields of the index" (but we already limit ourselves to dimensions so doesn't matter to us).

Lastly if no condition was intercepted we will take a mariadb key scan request and convert it ranges. 

A key scan request can be something like WHERE dim1 = 1 and dim2 > 10  basically what we already have supported for pushdowns. The difference is sometimes mariadb wants to use key scans so it can enforce unique/sorting of results.

the difference in the DDL:
```
show create table zscore_factor1_local\G
*************************** 1. row ***************************
       Table: test1
Create Table: CREATE TABLE `test1` (
  `ID` int(11) NOT NULL `dimension`=1 `lower_bound`='0' `upper_bound`='250000' `tile_extent`='1',
  `DATE` int(11) NOT NULL `dimension`=1 `lower_bound`='0' `upper_bound`='22001231' `tile_extent`='7300',
  `attr1` double DEFAULT NULL,
  `attr2` text DEFAULT NULL,
  `attr3` text DEFAULT NULL,
  `attr4` text DEFAULT NULL,
  PRIMARY KEY (`ID`,`DATE`)
) ENGINE=MyTile DEFAULT CHARSET=latin1 `uri`='file:///home/seth/tiledb_arrays/test1' `array_type`='SPARSE' `capacity`=1000000 `cell_order`=ROW_MAJOR `tile_order`=ROW_MAJOR
1 row in set (0.000 sec)
MariaDB [test]> show create table zscore_factor1_local_no_keys\G
*************************** 1. row ***************************
       Table: test1_no_keys
Create Table: CREATE TABLE `test1_no_keys` (
  `ID` int(11) DEFAULT NULL `dimension`=1 `lower_bound`='0' `upper_bound`='250000' `tile_extent`='1',
  `DATE` int(11) DEFAULT NULL `dimension`=1 `lower_bound`='0' `upper_bound`='22001231' `tile_extent`='7300',
  `attr2` double DEFAULT NULL,
  `attr3` text DEFAULT NULL,
  `attr4` text DEFAULT NULL,
  `attr5` text DEFAULT NULL
) ENGINE=MyTile DEFAULT CHARSET=latin1 `uri`='file:///home/seth/tiledb_arrays/test1' `array_type`='SPARSE' `capacity`=1000000 `cell_order`=ROW_MAJOR `tile_order`=ROW_MAJOR
1 row in set (0.001 sec)
```

Simply for "primary keys" we just list dimensions in order. The user does not have to do anything to enable this, I've got it on by default.

Since we don't always want to enforce primary key uniqueness, there is a new config option `mytile_dimensions_are_primary_keys` which controls the behavior.

Also introduced was `mytile_compute_table_records` for computing the estimated records on table open. Record count can be an important factor in the optimizer for things like joins and group bys. However TileDB does not currently have fast (10s of millisecond level) and efficient way to do this so we make it optional (disabled by default) to preserve performance for small queries.


Shelnutt2:
@normanb this is ready for review now. I will update docs once it is merged.

mytile/ha_mytile.h
```diff
@@ -235,6 +235,8 @@ class mytile : public handler {
 */
   const COND *cond_push(const COND *cond) override;
 
+  const COND *cond_push_local(const COND *cond);
```
joe-maley: Does this need a block comment?

mytile/ha_mytile.cc
```diff
@@ -974,6 +997,12 @@ void tile::mytile::cond_pop() {
   DBUG_VOID_RETURN;
 }
 
+Item *tile::mytile::idx_cond_push(uint keyno, Item *idx_cond) {
+  DBUG_ENTER("tile::mytile::idx_cond_push");
+  auto ret = cond_push_local(static_cast<Item_cond *>(idx_cond));
+  DBUG_RETURN(const_cast<Item *>(ret));
```
joe-maley: i'm not a huge fan of using `const_cast` because generally the right thing to do is to refactor the interface to return a non-const. but in this case, it looks like `cond_push_local` will always return either the input pointer, `idx_cond` or `nullptr`. Although more verbose, I would choose to do the following:
```
if (!cond_push_local(static_cast<Item_cond *>(idx_cond))) {
  return nullptr;
}
return idx_cond;
```

mytile/ha_mytile.cc
```diff
@@ -1551,6 +1581,128 @@ bool tile::mytile::valid_pushed_in_ranges() {
...
+
+  DBUG_RETURN(0);
+}
+
+int tile::mytile::index_end() {
+  DBUG_ENTER("tile::mytile::index_end");
+  DBUG_RETURN(rnd_end());
+}
+
+/**
```
joe-maley: should this block comment go in the header?

mytile/ha_mytile.h
```diff
@@ -291,6 +293,42 @@ class mytile : public handler {
    */
   int info(uint) override;
 
+  int index_read(uchar *buf, const uchar *key, uint key_len,
```
joe-maley: i think all of these routine should have their block comments here instead of in the `.cc` file.

mytile/ha_mytile.cc
```diff
@@ -1551,6 +1581,128 @@ bool tile::mytile::valid_pushed_in_ranges() {
...
+  DBUG_ENTER("tile::mytile::reset_pushdowns_for_key");
+  std::vector<std::shared_ptr<tile::range>> ranges_from_keys =
+      tile::build_ranges_from_key(key, key_len, find_flag,
+                                  this->array_schema->domain().type());
+
+  if (!ranges_from_keys.empty()) {
+    this->pushdown_ranges.clear();
+    this->pushdown_in_ranges.clear();
+
+    for (uint64_t i = 0; i < this->ndim; i++) {
```
joe-maley: Use this instead of the for loop:
```
this->pushdown_ranges.resize(this->ndim)
this->pushdown_in_ranges.resize(this->ndim)
```

mytile/ha_mytile.cc
```diff
@@ -1551,6 +1581,128 @@ bool tile::mytile::valid_pushed_in_ranges() {
...
+      this->pushdown_ranges.emplace_back();
+      this->pushdown_in_ranges.emplace_back();
+    }
+
+    // Copy shared pointer to main range pushdown
+    for (uint64_t i = 0; i < ranges_from_keys.size(); i++) {
+      this->pushdown_ranges[i].push_back(ranges_from_keys[i]);
```
joe-maley: You can avoid a copy with: 
`this->pushdown_ranges[i].emplace_back(std::move(ranges_from_keys[i]));`

Currently, each element is copied and the internal shared pointer ref counter is incremented to 2. When this routine exits, they're all decremented. The ref counters must be spinlocked so we save time there too.

This also would only work if `tile::range` has a move assignment operator, idk if it does.

joe-maley: In other words: ignore this if `tile::range` only has a copy-assignment operator.

mytile/mytile-range.cc
```diff
@@ -305,3 +305,86 @@ tile::get_unique_non_contained_in_ranges(
...
+    return Item_func::Functype::EQ_FUNC;
+  case HA_READ_MBR_CONTAIN:
+  case HA_READ_MBR_INTERSECT:
+  case HA_READ_MBR_WITHIN:
+  case HA_READ_MBR_DISJOINT:
+  case HA_READ_MBR_EQUAL:
+    my_printf_error(ER_UNKNOWN_ERROR, "Unsupported ha_rkey_function",
+                    ME_ERROR_LOG | ME_FATAL);
+  }
+  return Item_func::Functype::EQ_FUNC;
```
joe-maley: add a `default` or some compilers might complain

mytile/mytile-range.h
```diff
@@ -330,4 +330,138 @@ std::vector<std::shared_ptr<range>> get_unique_non_contained_in_ranges(
...
+    return {};
+
+  uint dims_with_keys = length / sizeof(T);
+
+  std::vector<std::shared_ptr<tile::range>> ranges;
+
+  // Cast key to array of type T
+  const T *key_typed = reinterpret_cast<const T *>(key);
+  // Handle all keys but the last one
+  for (size_t i = 0; i < dims_with_keys - 1; i++) {
```
joe-maley: prefer `++i` to `i++`. It doesn't really matter for primitives but allows for easier optimizations in the compiler for classes. Either way, modern convention is to prefer the pre-increment operator.

mytile/mytile-range.h
```diff
@@ -330,4 +330,138 @@ std::vector<std::shared_ptr<range>> get_unique_non_contained_in_ranges(
...
+                        std::malloc(sizeof(T)), &std::free),
+                    Item_func::EQ_FUNC, datatype});
+
+    // Copy value into range settings
+    // We can set lower/upper to key value because enforces equality for all key
+    // parts except the last that is set which is handled below
+    memcpy(range->lower_value.get(), &key_typed[i], sizeof(T));
+    memcpy(range->upper_value.get(), &key_typed[i], sizeof(T));
+
+    ranges.push_back(std::move(range));
```
joe-maley: prefer `emplace_back` to `push_back`, but this is ok

mytile/ha_mytile.cc
Shelnutt2: Does resize infer the emplace of the interior vectors, so they are valid vectors and not null?

mytile/mytile-range.h
Shelnutt2: emplace_back creates the container right? In this case I want to move the range to live on the vector, so push_back allows that to happen without a copy?

mytile/ha_mytile.cc
Shelnutt2: `tile::range` is just a plain struct. https://github.com/TileDB-Inc/TileDB-MariaDB/blob/a07e5702031154097002cdcf3df606aa1ffec725/mytile/mytile-range.h#L57

mytile/ha_mytile.cc
Shelnutt2: I need to collect the returned condition because while I'm not suppose to modify the existing condition I might return a new condition. In the case of `idx_cond_push`, its likely I'll either pushdown the whole condition or not at all like you say BUT I'm allowed to accept pushdown of only partial conditions. I'd like to leave the ability here so when we advance our pushdown support this will inherit the better support.

Also the reason for the const casting is because the main `cond_push` passes a const as we are not allowed to alter it. There is slighly different contracts between condition pushdown and index condition pushdown but for TileDB since we can only pushdown dimensions, we treat them the same (`cond_push_local`). To use one function I have to `const_cast` somewhere, since the return times vary.

https://github.com/MariaDB/server/blob/088de81d965f1b5ae2f50b1ed6e7cfccc0366098/sql/handler.h#L4207

https://github.com/MariaDB/server/blob/088de81d965f1b5ae2f50b1ed6e7cfccc0366098/sql/handler.h#L4140

PR #56, opened 2019-12-26
 Add support for encrypted arrays

Shelnutt2:
Support for creating, reading and writing encrypted arrays is included.

This includes minor modifications to create_array and discover_array in order to allow passing the encryption string to the handler.

Due to a limitation of MariaDB if you pass an encryption key it will be shown in the create table statement.

This will be rebased after #54 and #53 are merged.

### DIFF
--- a/mytile/ha_mytile.cc
+++ b/mytile/ha_mytile.cc
@@ -64,6 +64,7 @@ ha_create_table_option mytile_table_option_list[] = {
     HA_TOPTION_ENUM("tile_order", tile_order, "ROW_MAJOR,COLUMN_MAJOR",
                     TILEDB_ROW_MAJOR),
     HA_TOPTION_NUMBER("open_at", open_at, UINT64_MAX, 0, UINT64_MAX, 1),
+    HA_TOPTION_STRING("encryption_key", encryption_key),
     HA_TOPTION_END};
 
 // Structure for specific field options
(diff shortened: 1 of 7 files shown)


mytile/ha_mytile.cc
```diff
@@ -188,12 +190,25 @@ int tile::mytile::open(const char *name, int mode, uint test_if_locked) {
 
   // Open TileDB Array
   try {
+    std::string encryption_key;
+    if (this->table->s->option_struct->encryption_key != nullptr) {
+      sql_print_information("inside table->s");
```
joe-maley: this looks like a debug statement, remove? same for line 199

mytile/ha_mytile.cc
```diff
@@ -188,12 +190,25 @@ int tile::mytile::open(const char *name, int mode, uint test_if_locked) {
...
+      encryption_key = this->table->s->option_struct->encryption_key;
+    }
+    if (this->table_share->option_struct->encryption_key != nullptr) {
```
joe-maley: it appears that when this conditional is true, it may override the assignment on line 196. If this is intentional, could we put lines 194-197 as an 'if else' from this 'if' statement?

mytile/ha_mytile.cc
```diff
@@ -359,9 +373,17 @@ int tile::mytile::create_array(const char *name, TABLE *table_arg,
   }
 
   try {
+    std::string encryption_key;
+    if (create_info->option_struct->encryption_key != nullptr) {
+      encryption_key = create_info->option_struct->encryption_key;
```
joe-maley: on line 377 we compare `...->encryption_key` as a pointer, we should use something like `encryption_key = std::string(...->encryption_key)` to make the string constructor explicit. this is just style, no perf

joe-maley: same on similar lines in this patch

mytile/ha_mytile.h
```diff
@@ -536,5 +536,9 @@ class mytile : public handler {
    * @return true if query is complete, false otherwise
    */
   bool query_complete();
+
+  // Indicator for when non_empty_domain is empty and reads will return empty
+  // dataset
+  int empty_read = 0;
```
joe-maley: was there a reason to not use a `bool` here? more importantly, let's declare this in the .cc file so we don't expose it to every file that imports this header file.

PR #59, opened 2019-12-29
Add support for querying Array Metadata

Shelnutt2:
Metadata can be accessed via adding `@metadata` to the end of the URI.

The metadata is loaded into a std::unordered_map internally on every array open. This is then returned in two columns `key` and `value`.

If the value is an array, it is converted to a csv string for listing.

```
select * from `/my/array@metadata`
key	value
key4	25.1,26.2,27.3,28.4
key3	25.1
key1	25
key5	This is TileDB array metadata
key2	25,26,27,28
```

or
```
create table t1 uri='s3://my_bucket/my_array@metadata'
select * from t1;
```

```
select * from `/my/array@metadata` WHERE `key`='key' AND `value`='test';
```


Shelnutt2:
@stavrospapadopoulos would like your feedback on the `@metadata` scheme. If we want dynamic table discovery, `select * from /my/array@metadata` we have to do something in the uri to trigger whether its the array or metadata we want to query. If we say that dynamic querying doesn't matter and a user has to run a `CREATE TABLE` statement first then that opens a little more possibilities if you don't like the `@metadata`.

mytile/mytile-metadata.cc
```diff
@@ -0,0 +1,221 @@
...
+    tiledb_datatype_t type;
+    uint32_t num;
+    array->get_metadata_from_index(i, &key, &type, &num, &data);
+    std::string value = build_metadata_value_string(thd, data, num, type);
+
+    if (key.length() > *longest_key) {
+      *longest_key = key.length();
+    }
+
+    metadata_map.emplace(key, value);
```
joe-maley: since we're done with `key` and `value` here,  you could `std::move()` them to avoid copies

mytile/mytile-metadata.cc
```diff
@@ -0,0 +1,221 @@
...
+  return str;
+}
+
+std::unordered_map<std::string, std::string>
+tile::build_metadata_map(THD *thd, std::shared_ptr<tiledb::Array> array,
+                         uint64_t *longest_key) {
+  std::unordered_map<std::string, std::string> metadata_map;
+  *longest_key = 0;
+
+  for (uint64_t i = 0; i < array->metadata_num(); i++) {
```
joe-maley: prefer `++i` for consistency, even though its a primitive and the performance won't matter

mytile/mytile-metadata.h
```diff
@@ -0,0 +1,125 @@
...
+template <typename T>
+std::string build_metadata_numeric_value_string(const void *data,
+                                                uint32_t num) {
+  std::stringstream ss;
+  const T *data_typed = static_cast<const T *>(data);
+  for (uint32_t i = 0; i < num; i++) {
+    ss << data_typed[i] << METADATA_DELIMITER;
+  }
+  std::string s = ss.str();
+  return s.substr(0, s.length() - 1);
```
joe-maley: same comment as earlier: include a code comment to indicate what the `- 1` is doing

mytile/mytile-discovery.cc
```diff
@@ -297,3 +339,101 @@ int tile::discover_array(THD *thd, TABLE_SHARE *ts, HA_CREATE_INFO *info) {
...
+    }
+
+    if (schema->cell_order() == tiledb_layout_t::TILEDB_ROW_MAJOR) {
+      table_options << " cell_order=ROW_MAJOR";
+    } else if (schema->cell_order() == tiledb_layout_t::TILEDB_COL_MAJOR) {
+      table_options << " cell_order=COL_MAJOR";
+    } else {
+      const char *layout;
+      tiledb_layout_to_str(schema->cell_order(), &layout);
+      throw std::runtime_error(
```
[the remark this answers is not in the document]
Shelnutt2: The previous two lines are so we can get the string name of what isn't supported. 99% this will be unknown/not valid but if we add a new layout then this would print the nice message saying we need to go back to this if statements and add support for what is missing.

PR #69, opened 2020-05-02
Empty predicates for primary keys return wrong non empty domain

antalakas:


mytile/mytile-range.cc
```diff
@@ -667,6 +667,10 @@ tile::build_ranges_from_key(const uchar *key, uint length,
       last_key = true;
     }
 
+    if ((key + key_offset)[0] == '\0') {
```
Shelnutt2, line 670: I think this needs to be a `if (key_offset >= key_length) break;` and we can put it at the top of the loop. We are not breaking out when we have run out of the keys, good catch on this!

PR #71, opened 2020-05-11
Fix heterogeneus dims for mrr

antalakas:


mytile/mytile-range.cc
```diff
@@ -230,6 +230,9 @@ std::shared_ptr<tile::range> tile::merge_ranges_to_super(
   case tiledb_datatype_t::TILEDB_UINT64:
     return merge_ranges_to_super<uint64_t>(ranges);
 
+  case tiledb_datatype_t::TILEDB_STRING_ASCII:
+    return merge_ranges<char>(ranges);
```
Shelnutt2, line 234: Will this work? Is `merge_ranges` compatible with the `merge_ranges_to_super` functionality? Or do you need a new `merge_ranges_to_super_str` like `merge_ranges_str` ?

mytile/ha_mytile.cc
```diff
@@ -2202,20 +2205,18 @@ int tile::mytile::build_mrr_ranges() {
...
   while (!mrr_funcs.next(mrr_iter, &mrr_cur_range)) {
 
     if (mrr_cur_range.start_key.key != nullptr) {
-      for (uint64_t i = 0; i < mrr_cur_range.start_key.length / datatype_size;
-           i++) {
+      for (uint64_t i = 0; i < this->ndim; i++) {
         auto &range = tmp_ranges[i][0];
         update_range_from_key_for_super_range(range, mrr_cur_range.start_key, i,
-                                              datatype);
+                                              dims[i].type());
```
Shelnutt2, line 2219: update_range_from_key_for_super_range has to be adjusted because it currently loops over every key. You will need to adjust update_range_from_key_for_super_range to only work on a single range, single dim and single key. You will have to use a key_offset like `build_ranges_from_key`.

mytile/mytile-range.cc
antalakas: It seems that we never call `merge_ranges_str`

PR #79, opened 2020-06-02
Improve mrr path filtering

antalakas:
`primary_key` and `mrr_incomplete_queries` tests do not pass


mytile/ha_mytile.cc
```diff
@@ -2032,10 +2032,10 @@ int tile::mytile::index_read_scan(const uchar *key, uint key_len,
            (find_flag == ha_rkey_function::HA_READ_KEY_EXACT ||
             find_flag == ha_rkey_function::HA_READ_KEY_OR_NEXT ||
             find_flag == ha_rkey_function::HA_READ_KEY_OR_PREV)) ||
-          (key_cmp < 0 && (find_flag == ha_rkey_function::HA_READ_BEFORE_KEY ||
+          (key_cmp > 0 && (find_flag == ha_rkey_function::HA_READ_BEFORE_KEY ||
```
Shelnutt2, line 2035: This change is not correct, `compare_key_to_dims` need to be adjusted so the proper values are returned then the greater than and less than comparison will be valid.

antalakas: `key_cmp < 0 ` means that the key value for the corresponding dimension is less than the value of the record in the key field. So, actually in his case the combination of the function needed to pass is `HA_READ_AFTER_KEY` or `HA_READ_KEY_OR_NEXT` because we expect the record field value to indeed be greater than the key

PR #86, opened 2020-07-22
Sets dimension fields to NOT NULL if array allows duplicates

antalakas:
In array discovery

Omitting the if statement leads to error in mrr test for `bank` array


mytile/mytile-discovery.cc
```diff
@@ -237,9 +237,14 @@ int tile::discover_array(THD *thd, TABLE_SHARE *ts, HA_CREATE_INFO *info) {
 
     for (const auto &dim : schema->domain().dimensions()) {
       int mysql_type = TileDBTypeToMysqlType(dim.type(), false);
+
       sql_string << std::endl
                  << "`" << dim.name() << "` " << MysqlTypeString(mysql_type);
 
+      if (schema->allows_dups()) {
```
Shelnutt2, line 244: We should set NOT NULL for all dimensions regardless of if dups are enabled or not.

PR #88, opened 2020-07-24
Incorrectly creating array on local disk instead of s3

antalakas:
If `tiledb://namespace` is not given, and token is set, the array is created in an s3 bucket but not registered with the console.

### DIFF
--- a/mytile/ha_mytile.cc
+++ b/mytile/ha_mytile.cc
@@ -406,6 +406,13 @@ int tile::mytile::create_array(const char *name, TABLE *table_arg,
 
     // Get array uri from name or table option
     std::string create_uri = name;
+
+    if( (strncmp(table_arg->s->table_name.str, "s3://", 5) == 0) ||
+        (strncmp(table_arg->s->table_name.str, "azure://", 8) == 0) ||
+        (strncmp(table_arg->s->table_name.str, "gcs://", 6) == 0) ||
+        (strncmp(table_arg->s->table_name.str, "tiledb://", 9) == 0))
+      create_uri = table_arg->s->table_name.str;
+
     if (create_info->option_struct->array_uri != nullptr)
       create_uri = create_info->option_struct->array_uri;
 


mytile/ha_mytile.cc
```diff
@@ -405,7 +405,7 @@ int tile::mytile::create_array(const char *name, TABLE *table_arg,
     }
 
     // Get array uri from name or table option
-    std::string create_uri = name;
```
Shelnutt2, line 408: Can you check where we get the name from? We might want to fix this further up in the call stack

antalakas: It comes from `create function` which us a pure virtual function and the signature is needed as is to get called by the engine

antalakas: I have issues with the debugger (exits the debug process on breakpoint hit) so could not go up the stack, i just removed the name param from everywhere and the resulting function signature (create) shadowed the abstract implementation (less ags) so this was not the right thing to do

Shelnutt2: Thanks for clarifying.

mytile/ha_mytile.cc
```diff
@@ -406,6 +406,10 @@ int tile::mytile::create_array(const char *name, TABLE *table_arg,
 
     // Get array uri from name or table option
     std::string create_uri = name;
+
+    if(strstr(table_arg->s->table_name.str, "s3://") != NULL)
```
Shelnutt2, line 410: Use tiledb's vfs function for identifying if this is a local file or a remote one. We need to support all possible remotes, `s3://`, `azure://`, `gcs://`, `tiledb://`.

antalakas: Could not find a relevant function

mytile/ha_mytile.cc
```diff
@@ -406,6 +406,13 @@ int tile::mytile::create_array(const char *name, TABLE *table_arg,
 
     // Get array uri from name or table option
     std::string create_uri = name;
+
+    if( (strstr(table_arg->s->table_name.str, "s3://") != NULL) &&
```
Shelnutt2, line 410: It is probably safer to use `strncmp`

antalakas: 👍

mytile/ha_mytile.cc
```diff
@@ -407,10 +407,10 @@ int tile::mytile::create_array(const char *name, TABLE *table_arg,
     // Get array uri from name or table option
     std::string create_uri = name;
 
-    if( (strstr(table_arg->s->table_name.str, "s3://") != NULL) &&
-        (strstr(table_arg->s->table_name.str, "azure://") != NULL) &&
-        (strstr(table_arg->s->table_name.str, "gcs://") != NULL) &&
-        (strstr(table_arg->s->table_name.str, "tiledb://") != NULL))
+    if( (strcmp(table_arg->s->table_name.str, "s3://") == 0) ||
```
Shelnutt2, line 410: you want `strncmp`, where there is a 3rd parameter of the number of characters to compare.

PR #95, opened 2020-09-10
Update plugin maturity to beta

Shelnutt2:
Update plugin maturity to beta

### DIFF
--- a/mytile/ha_mytile.cc
+++ b/mytile/ha_mytile.cc
@@ -2492,7 +2492,7 @@ mysql_declare_plugin(mytile){
     PLUGIN_LICENSE_PROPRIETARY, /* the plugin license (PLUGIN_LICENSE_XXX) */
     mytile_init_func,           /* Plugin Init */
     NULL,                       /* Plugin Deinit */
-    0x0003,                     /* version number (0.3) */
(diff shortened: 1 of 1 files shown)


mytile/ha_mytile.cc
```diff
@@ -2492,7 +2492,7 @@ mysql_declare_plugin(mytile){
     PLUGIN_LICENSE_PROPRIETARY, /* the plugin license (PLUGIN_LICENSE_XXX) */
     mytile_init_func,           /* Plugin Init */
     NULL,                       /* Plugin Deinit */
-    0x0003,                     /* version number (0.3) */
+    0x0059,                     /* version number (0.3) */
```
antalakas, line 2495: Comment should be `/* version number (0.5.9) */` ?

antalakas (approved):

mytile/ha_mytile.cc
Shelnutt2: Fixed, and I bumped the version to 0.6.0. I'm going to look to rebase on 10.5 over the next day or two and we'll cut a new release as 0.6. Or if not we'll just cut a 0.6 with the fixes we have since 0.5.9.

antalakas: 👍

PR #96, opened 2020-09-11
Update for MariaDB 10.5

Shelnutt2:
Set MariaDB version to 10.5 for CI/Docker, also update to handle removal of `primary_key_is_clustered` function, new parameter to `records_in_ranges`, remove of `state` attribute on handler struct and lastly rename of `utf8` to `utf8mb3`.

There are no functional changes in any of these updates.


mytile/ha_mytile.cc
```diff
@@ -2241,8 +2240,9 @@ int tile::mytile::index_read_idx_map(uchar *buf, uint idx, const uchar *key,
   DBUG_RETURN(index_read_scan(key, key_len, find_flag));
 }
 
-ha_rows tile::mytile::records_in_range(uint inx, key_range *min_key,
-                                       key_range *max_key) {
+ha_rows tile::mytile::records_in_range(uint inx, const key_range *min_key,
+                                       const key_range *max_key,
+                                       page_range *page) {
```
antalakas, line 2245: What is the purpose of this param?

antalakas (approved):

mytile/ha_mytile.cc
Shelnutt2: See: https://github.com/MariaDB/server/commit/f36ca142f7fa59598e34e842b60372b963067766

PR #98, opened 2020-10-13
Erica/ch3335/tiledb 2 1

ericawithlove208:
This updates the TileDB version used in MyTile from version 2.0.8 to 2.1.0.

Support is added for attribute fill in values. These are implemented through DEFAULT clauses within CREATE TABLE statements. Any attribute (ie column not marked as a dimension) with either an implicit or explicit DEFAULT clause will have that default value supplied to Attribute::set_fill_value(...). 

As a result, we no longer require all columns to be specified within an INSERT statement. 

Arrays created outside of Maria will also have attribute fill values registered as DEFAULT values within Maria during table discovery.


mysql-test/mytile/r/fill_in.result
```diff
@@ -0,0 +1,45 @@
+# int varchar and double defaults
+CREATE TABLE t1 (
+dim0 integer DIMENSION=1 lower_bound="0" upper_bound="100" tile_extent="10",
+dim1 integer DIMENSION=1 lower_bound="0" upper_bound="100" tile_extent="10",
+attr0 integer DEFAULT 12345 NOT NULL,
```
Shelnutt2, line 5: This is great!

Shelnutt2 (commented):
This looks pretty good!

Can you also add support for default/fill values in the table discovery? This will allow for "show create table" statements to display and for existing arrays to properly register their defaults with mariadb: https://github.com/TileDB-Inc/TileDB-MariaDB/blob/71e4206f9f9e0ae964df9d684002dcd69c5f2005/mytile/mytile-discovery.cc#L275

Shelnutt2 (approved):
lgtm@

Shelnutt2 (changes requested):
Forgot about the need to add discovery table support as we discussed, I'll hold off on the merge.

Shelnutt2:
@ericawithlove208 can you update to TileDB 2.1.1? Please update the dockerfiles also.

Shelnutt2 (approved):
Looks good! After you bump to 2.1.1 and add the dockerfiles LGTM

PR #101, opened 2020-10-20
Erica/ch3515/time

ericawithlove208:
This fixes an error when converting a time column (no date part) to seconds on insert. The failing mysql-test was added back into the run.

### DIFF
--- a/mytile/mytile.cc
+++ b/mytile/mytile.cc
@@ -951,10 +951,17 @@ int tile::set_buffer_from_field(Field *field, std::shared_ptr<buffer> &buff,
     MYSQL_TIME mysql_time;
     field->get_date(&mysql_time, date_mode_t(0));
 
-    // Convert time with timezone consideration
-    uint32_t not_used;
-    my_time_t seconds =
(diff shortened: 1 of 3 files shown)


mytile/mytile.cc
```diff
@@ -951,10 +951,17 @@ int tile::set_buffer_from_field(Field *field, std::shared_ptr<buffer> &buff,
     MYSQL_TIME mysql_time;
     field->get_date(&mysql_time, date_mode_t(0));
 
-    // Convert time with timezone consideration
-    uint32_t not_used;
-    my_time_t seconds =
-        thd->variables.time_zone->TIME_to_gmt_sec(&mysql_time, &not_used);
+    my_time_t seconds = 0;
+    // if we only have the time part
```
Shelnutt2, line 955: Is this needed for any other datetime types or is `TIME` always translated to NS?

ericawithlove208: it is only needed for time. The issue is that TIME_to_gmt_sec requires a valid date part. Each of the other DateTime and TypeStamp types require a valid date part before they are given to us, so they should be fine.

Shelnutt2 (approved):

PR #104, opened 2020-10-27
Condition pushdown support for DateTime types

ericawithlove208:
This adds support for condition pushdown for all TileDB DateTime datatypes.
Some conversions between Maria and TileDB types were fixed.
DateTime value representation between storage and ranges was unified.
Tests were added for all DateTime data types via the datetime_dimension array dataset.


mytile/mytile.cc
```diff
@@ -381,6 +407,82 @@ bool tile::MysqlDatetimeType(enum_field_types type) {
   return false;
 }
 
+int64_t tile::MysqlTimeToTileDBTimeVal(THD* thd, const MYSQL_TIME &mysql_time,
+                                       tiledb_datatype_t datatype) {
+  if (datatype == tiledb_datatype_t::TILEDB_DATETIME_YEAR  ||
+      datatype == tiledb_datatype_t::TILEDB_DATETIME_MONTH ||
+      datatype == tiledb_datatype_t::TILEDB_DATETIME_WEEK  ||
+      datatype == tiledb_datatype_t::TILEDB_DATETIME_DAY) {
+    switch(datatype) {
```
Shelnutt2, line 416: Would this be better as 4 if/if else statements instead of a sub switch?

Shelnutt2 (approved):
One minor comment but otherwise great job! This is is excellent and your test cases a very verbose, I like it!

PR #117, opened 2020-12-06
Add support for subsetting an array on discovery

Shelnutt2:
If a TileDB array has more than the 4096 max columns that MariaDB supports we want to allow the user to issues a create table statement to register a subset of the dimensions/attributes. This behavior is controlled by the new `mytile_create_allow_subset_existing_array` session parameter. The default behavior, error if issuing a create table with columns on an existing, is left intact as for most users they will not want subsetting but instead likely would be trying to create a
new array.

Example SQL usage on an array which has 5k attributes:
```
MariaDB [test]> create table test_1 engine=MyTile uri='/tmp/attributes_5k';
ERROR 1939 (HY000): Engine MyTile failed to discover table `test`.`test_1` with 'create table `test_1` (
`d0` INTEGER dimension=1 lower_bound='0' upper_bound='10000000' tile_extent='100',
`a4998` INTEGER DEFAULT -2147483648,
`a4997` INTEGER DEFAULT -2147483648,
`a4996` INTEGER DEFAULT -2147483648,
`a4991` INTEGER DEFAULT -2147483648,
`a4987` INTEGER DEFAULT -2147483648,
`a4986` INTEGER DEFAULT -2147483648,
`a4981` INTEGER DEFAULT -2147483648,
`a4980` INTEGER DEFAULT -2147483648,
`a4979` INTEGER DEFAULT -2147483648,
`a4977` IN
MariaDB [test]> set mytile_create_allow_subset_existing_array=1;
Query OK, 0 rows affected (0.000 sec)

MariaDB [test]> create table test_1 (d0 int, a0 int, a1 int, a2 int, a3 int)  engine=MyTile uri='/tmp/attributes_5k';
Query OK, 0 rows affected (0.012 sec)

MariaDB [test]> show create table test_1\G
*************************** 1. row ***************************
       Table: test_1
Create Table: CREATE TABLE `test_1` (
  `d0` int(11) DEFAULT NULL,
  `a0` int(11) DEFAULT NULL,
  `a1` int(11) DEFAULT NULL,
  `a2` int(11) DEFAULT NULL,
  `a3` int(11) DEFAULT NULL
) ENGINE=MyTile DEFAULT CHARSET=utf8 `uri`='/tmp/attributes_5k'
1 row in set (0.017 sec)

# Table is empty so no results expected but querying works
MariaDB [test]> select * from test_1;
Empty set (0.032 sec)
````

### DIFF
--- a/mysql-test/mytile/r/create_allow_subset_existing_array.result
+++ b/mysql-test/mytile/r/create_allow_subset_existing_array.result
@@ -1,0 +1,49 @@
+#
+# The purpose of this test is to subset the columns on an existing array
+#
+call mtr.add_suppression("\\[ERROR\\] (mysqld|mariadbd): Error in creating table \\[TileDB::ArraySchema\\] Error: Cannot set domain; Domain must contain at least one dimension");
+CREATE TABLE t1 (a int) ENGINE=mytile uri='MTR_SUITE_DIR/test_data/tiledb_arrays/1.6/quickstart_dense';;
+ERROR HY000: Error in creating table [TileDB::ArraySchema] Error: Cannot set domain; Domain must contain at least one dimension
+SET mytile_delete_arrays=0;
+SET mytile_create_allow_subset_existing_array=1;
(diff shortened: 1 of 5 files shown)


antalakas (approved):
nice!

PR #121, opened 2020-12-07
Nullable support

Shelnutt2:
This PR add support for nullable attributes in TileDB.


ericawithlove208 (approved):

mytile/mytile.h
```diff
@@ -296,16 +394,49 @@ int set_string_buffer_from_field(Field *field, std::shared_ptr<buffer> &buff,
...
+      return ERR_WRITE_FLUSH_NEEDED;
+    }
+    if (field_null) {
+      // XXX : zero length single cell writes are not supported (write some
+      // trash)
+      if (res->length() == 0) {
+        memcpy(static_cast<T *>(buff->buffer) + start, "0", 1);
+        buff->buffer_size += 1;
+      }
+      buff->validity_buffer[i] = (uint8_t)0;
```
Shelnutt2, line 419: Let's use c++ style casting, `static_cast<uint8_t>(0)` here and the other places too if you don't mind.

ericawithlove208: ah yes, will do.

PR #125, opened 2020-12-14
Add support for time traveling with <uri>@<time>

Shelnutt2:
This allows a secondary form of time traveling by appending @<timestamp> to the array URI. This enabled time travling with dynamic discovery without the needed for a create table statement.

Usage:

```
select * from `s3://mybucket/myarray@1`
```

This is equivalent to the previous way:
```
CREATE TABLE myarray engine=MyTile open_at=1;
select * from myarray;
```


antalakas (approved):
lgtm

PR #129, opened 2020-12-21
Add test arrays to dockerfile min

antalakas:


Shelnutt2:
@antalakas how large are the arrays?

antalakas:
> @antalakas how large are the arrays?

5MB based on the diff in image size

Shelnutt2 (approved):

PR #131, opened 2020-12-23
Add Dockerfile min to CI

antalakas:


azure-pipelines.yml
```diff
@@ -43,6 +43,10 @@ stages:
             dockerfile: docker/Dockerfile
             repository: tiledb/tiledb-mariadb
             context: .
+          tiledb-mariadb-min:
+            dockerfile: docker/Dockerfile-min
+            repository: tiledb/tiledb-mariadb
```
Shelnutt2, line 48: Please update the upstream name. This is were the image is pushed to.

Shelnutt2 (changes requested):

azure-pipelines.yml
antalakas: 👍

Shelnutt2 (approved):

PR #134, opened 2021-01-21
Added mariadb.sys user to the list of system users to fix Docker entrypoint

sgloutnikov:
Launching the `tiledb/tiledb-mariadb` and `tiledb/tiledb-mariadb-server` docker images currently result in an infinite loop waiting for MariaDB to initialize due to the following error: `The user specified as a definer ('mariadb.sys'@'localhost') does not exist`.

This change adds the `mariadb.sys` user to the exclusion list. More information on the MariaDB changes that caused this problem can be found in the comments of [this ticket](https://jira.mariadb.org/browse/MDEV-22542).

### DIFF
--- a/docker/docker-entrypoint-interactive.sh
+++ b/docker/docker-entrypoint-interactive.sh
@@ -167,7 +167,7 @@ EOSQL
     --  or products like mysql-fabric won't work
     SET @@SESSION.SQL_LOG_BIN=0;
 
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):
Thank you for this PR, great catch on this. We plan to get a new release out today or tomorrow with the #121 and this will be included. I will circle back and comment on this PR to let you know when the updated docker image is out.

PR #137, opened 2021-02-08
conditionally include plugin load add for mytile based on embedded build

ericawithlove208:
adds a confugure_file step to CMakeLists which enables/disables mytile plugin loading in my.cnf depending on the value of WITH_EMBEDDED_SERVER.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,3 +126,11 @@ if(NOT TileDB_FOUND AND SUPERBUILD)
     target_include_directories(mytile PUBLIC "${EP_INSTALL_PREFIX}/include" "mytile")
     INSTALL(FILES ${TILEDB_INSTALL_LIBRARIES} DESTINATION ${INSTALL_PLUGINDIR})
 endif()
+
+# Set my.cnf values based on build flags
+if(WITH_EMBEDDED_SERVER)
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):

CMakeLists.txt
```diff
@@ -126,3 +126,11 @@ if(NOT TileDB_FOUND AND SUPERBUILD)
     target_include_directories(mytile PUBLIC "${EP_INSTALL_PREFIX}/include" "mytile")
     INSTALL(FILES ${TILEDB_INSTALL_LIBRARIES} DESTINATION ${INSTALL_PLUGINDIR})
 endif()
+
+# Set my.cnf values based on build flags
+if(WITH_EMBEDDED_SERVER)
+    SET(MYTILE_PLUGIN_LOAD_ADD "# plugin-load-add=ha_mytile")
```
Shelnutt2, line 132: You could even make this an empty string instead of commenting it out.

ericawithlove208: I like it explicitly commented out as a reminder of what mode we're in - can change if you'd like.

Shelnutt2: No this works for me!

PR #140, opened 2021-03-17
Fix handling of string key comparison and ranges

Shelnutt2:
This fixes two issues, the first being that we incorrectly were copying the key for pushdown onto the TileDB range. We were accidentally only considering the first character instead of all characters.

Second we were not handling partial key matches. For strings a partial key comparison might include zero length keys, which should be considered for any match. This adjusts to capture the comparison of zero length string.

### DIFF
--- a/mytile/mytile-range.h
+++ b/mytile/mytile-range.h
@@ -568,7 +568,11 @@ build_range_from_key(const uchar *key, uint length,
     return {};
 
   // Cast key to array of type T
-  const T *key_typed = reinterpret_cast<const T *>(key);
+  // Ideally we'd avoid a copy here but we are unlikely to have a large number
+  // of keys or keys that are large in size so the copying is okay for
+  // simplicity of the code
+  T *key_typed = static_cast<T *>(std::malloc(size));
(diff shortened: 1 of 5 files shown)


antalakas (approved):
lgtm

mytile/mytile-range.h
```diff
@@ -590,16 +591,16 @@ build_range_from_key(const uchar *key, uint length,
   case Item_func::GT_FUNC: {
     range->operation_type = Item_func::GE_FUNC;
     if (std::is_floating_point<T>()) {
-      lower = std::nextafter(lower, std::numeric_limits<T>::max());
+      *key_typed = std::nextafter(*key_typed, std::numeric_limits<T>::max());
     } else if (std::is_arithmetic<T>()) {
-      lower += 1;
+      *key_typed += 1;
```
ihnorton, line 596: is there any risk of overflow here?

Shelnutt2: Yes there is, I'll add a story to refactor to handle this. Specifically its applicable only to the integer datatypes.

PR #143, opened 2021-03-18
index_next_same should conditional reset indexes

Shelnutt2:
index_next_same should conditional reset indexes

### DIFF
--- a/mytile/ha_mytile.cc
+++ b/mytile/ha_mytile.cc
@@ -2881,7 +2881,7 @@ int tile::mytile::multi_range_read_explain_info(uint mrr_mode, char *str,
 
 int tile::mytile::index_next_same(uchar *buf, const uchar *key, uint keylen) {
   int error;
-  DBUG_ENTER("handler::index_next_same");
+  DBUG_ENTER("mytile::index_next_same");
   // Store the current indexes so we can reset them after checking
   // if the next key is the same
   uint64_t record_index_before_check = this->record_index;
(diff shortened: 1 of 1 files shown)


mytile/ha_mytile.cc
```diff
@@ -2923,14 +2923,15 @@ int tile::mytile::index_next_same(uchar *buf, const uchar *key, uint keylen) {
       table->record[0] = save_record_0;
       for (key_part = key_info->key_part; key_part < key_part_end; key_part++)
         key_part->field->move_field_offset(-ptrdiff);
+
+      // Reset the record indexes back to before the first read
```
ihnorton, line 2927: Where is record_index modified such that it needs to be reset here?

Shelnutt2: Its modified in `index_next` -> `scan_rnd_row`.

Sometime with indexes, and/or MRR queries MariaDB will ask for the next record to see if the keys/index fields are identical or not. If they are different, then MariaDB will stop collecting records and performance what batch operation it is doing (join/groupby/etc). Later it will then call `index_next` assuming it will get the next key.

Imagine a sequence of records:
```
DIM1
-------
1
1
2
```

MariaDB in this code path will read `1`, then it calls `index_next_same`, gets `1` again, add this to its internal buffer, calls `index_next_same` again, gets `2`. It sees `2` is different than `1`, so it doesn't use the record, stops does it functional operation. Later when it resumes the read, it will call `index_next` and expects to get `2`.

ihnorton (approved):

PR #144, opened 2021-03-18
ref for position can't be reset with MRR

Shelnutt2:
For variable length dimensions we were resetting the `ref_length` and reallocating `ref` when the handler called for `position()` based on the string size of the current coordinates. This doesn't work for MRR cases though because the address of `ref` is taken before `position()` is called.

The quick fix is we'll always set the `ref_length` to the maximum allowed size of a row. This means we'll allocate an extra 65KB, but in the grand scheme this is insignificant for most usages and we only generally store a small amount of positions.

### DIFF
--- a/mytile/ha_mytile.cc
+++ b/mytile/ha_mytile.cc
@@ -229,9 +229,18 @@ int tile::mytile::open(const char *name, int mode, uint test_if_locked) {
     // Set ref length used for storing reference in position(), this is the size
     // of a subarray for querying
     this->ref_length = 0;
+    bool any_var_length = false;
     for (const auto &dim : domain.dimensions()) {
+      if (dim.cell_val_num() == TILEDB_VAR_NUM) {
+        any_var_length = true;
+        break;
+      }
       this->ref_length += sizeof(uint64_t);
       this->ref_length += tiledb_datatype_size(dim.type());
+    }
+
+    if (any_var_length) {
+      this->ref_length = MAX_FIELD_VARCHARLENGTH;
     }
 
     // If the user requests we will compute the table records on opening the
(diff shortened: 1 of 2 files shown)


mytile/ha_mytile.cc
```diff
@@ -1207,7 +1199,11 @@ void tile::mytile::position(const uchar *record) {
...
   if (coords.size() > this->ref_length) {
-    realloc_ref_based_size(coords.size());
+    my_printf_error(
+        ER_UNKNOWN_ERROR,
+        "[position] error dimensions longer than ref_length for %s :",
+        ME_ERROR_LOG | ME_FATAL, this->uri.c_str());
+    DBUG_VOID_RETURN;
   }
 
   memcpy(this->ref, coords.data(), this->ref_length);
```
ihnorton, line 1209: Does `this->ref_length` need to get set to `coords.size()` before this call?

Shelnutt2: We set in the `open`. I have also seen some storage engines set it in `info`, but we only set it in `open`

ihnorton: What I'm trying to figure out is:
- with this change it's set to `this->ref_lenght = MAX_FIELD_VARCHARLENGTH`
- so it seems like we could have a situation where `coords.size() < MAX_FIELD_VARCHARLENGTH` -- but maybe I missed either where it is updated (or `coords` is guaranteed to be padded out to that length?)

ihnorton: (wondering if there could be a `memcpy` past the end of `coords)

Shelnutt2: This is a good catch, this will memcpy pasted the size of coords. I've now added a `std::min` to only the smallest of the two's size.

ihnorton (approved):
LGTM

PR #148, opened 2021-03-26
Handle datetime in MRR key merging

Shelnutt2:
Handle datetime in MRR key merging, this fixes an issue with a customer query where due to the size the condition was being applied as MRR.

### DIFF
--- a/mytile/mytile-range.cc
+++ b/mytile/mytile-range.cc
@@ -985,8 +985,10 @@ tile::build_ranges_from_key(THD *thd, const TABLE *table, const uchar *key,
 
 void tile::update_range_from_key_for_super_range(
     std::shared_ptr<tile::range> &range, key_range key, uint64_t key_offset,
-    const bool start_key, const bool last_key_part,
-    tiledb_datatype_t datatype) {
+    const bool start_key, const bool last_key_part, tiledb_datatype_t datatype,
+    THD *thd, Field *field) {
+  assert(thd);
+  assert(field);
   // Length shouldn't be zero here but better safe then segfault!
   if (key.length == 0)
     return;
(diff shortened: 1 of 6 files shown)


mytile/mytile-range.cc
```diff
@@ -985,8 +985,8 @@ tile::build_ranges_from_key(THD *thd, const TABLE *table, const uchar *key,
 
 void tile::update_range_from_key_for_super_range(
     std::shared_ptr<tile::range> &range, key_range key, uint64_t key_offset,
-    const bool start_key, const bool last_key_part,
-    tiledb_datatype_t datatype) {
+    const bool start_key, const bool last_key_part, tiledb_datatype_t datatype,
+    THD *thd, Field *field) {
```
ihnorton, line 989: add assert for these?

mytile/mytile-range.cc
```diff
@@ -1037,7 +1055,24 @@ void tile::update_range_from_key_for_super_range(
   case tiledb_datatype_t::TILEDB_DATETIME_NS:
   case tiledb_datatype_t::TILEDB_DATETIME_PS:
   case tiledb_datatype_t::TILEDB_DATETIME_FS:
-  case tiledb_datatype_t::TILEDB_DATETIME_AS:
+  case tiledb_datatype_t::TILEDB_DATETIME_AS: {
+    MYSQL_TIME mysql_time;
```
ihnorton, line 1059: do we need a dimension bounds check somewhere here?

ihnorton: (oh - is this why you posted the issue about auto clamping?)

ihnorton (approved):
I think I more-or-less follow what is happening here after reading some upstream docs...

LGTM, just minor suggestions/questions

mytile/mytile-range.cc
Shelnutt2: Yes, currently we are missing any bounds checks for the dimensions domain. I opened https://github.com/TileDB-Inc/TileDB/pull/2162 to add this support in core so we don't have to implement it on various applications.

PR #160, opened 2021-06-11
Add QueryCondition pushdown support

Shelnutt2:
This pushes down filtering on the attribute field. This is currently limited to AND clauses with basic predicates of `>`/`>=`/`=`/`<=`/`<`/`!=` along with `IS NULL` and `IS NOT NULL`.

At this time I opted to re-use the existing range building functionality we have. This minimized the changes needed to start with.

### DIFF
--- a/mytile/ha_mytile.cc
+++ b/mytile/ha_mytile.cc
@@ -304,6 +304,9 @@ int tile::mytile::close(void) {
     if (this->query != nullptr)
       this->query = nullptr;
 
+    if (this->query_condition != nullptr)
+      this->query_condition = nullptr;
+
     // close array
     if (this->array != nullptr && this->array->is_open())
       this->array->close();
@@ -691,6 +694,11 @@ int tile::mytile::init_scan(THD *thd) {
       }
     }
 
+    // If a query condition on an attribute was set, apply it
+    if (this->query_condition != nullptr) {
+      this->query->set_condition(*this->query_condition);
+    }
+
     if (!this->valid_pushed_ranges() &&
         !this->valid_pushed_in_ranges()) { // No pushdown
       if (this->empty_read)
@@ -1038,6 +1046,7 @@ int tile::mytile::rnd_end() {
   dealloc_buffers();
   this->pushdown_ranges.clear();
   this->pushdown_in_ranges.clear();
+  this->query_condition = nullptr;
   // Reset indicators
   this->record_index = 0;
   this->records = 0;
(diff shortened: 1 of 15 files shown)


Shelnutt2:
@ihnorton this is ready for review now.

mytile/ha_mytile.cc
```diff
@@ -1286,26 +1295,111 @@ const COND *tile::mytile::cond_push_func_datetime(const Item_func *func_item) {
...
 
   uint64_t dim_idx = 0;
-  tiledb_datatype_t dim_type = tiledb_datatype_t::TILEDB_ANY;
+  tiledb_datatype_t datatype = tiledb_datatype_t::TILEDB_ANY;
+
+  // Check attributes
+  bool nullable = false;
+  if (this->array_schema->has_attribute(column_field->field_name.str)) {
+
+    // Dense arrays do not support query conditions
```
ihnorton, line 1309: I've been wondering about this because there is a benchmark for QC which uses dense: https://github.com/TileDB-Inc/TileDB/blob/ea3f4c8aac36d66db4ff5167f5a8e2627a4e6601/test/benchmarking/src/bench_dense_attribute_filtering.cc#L84

mytile/ha_mytile.cc
```diff
@@ -1286,26 +1295,111 @@ const COND *tile::mytile::cond_push_func_datetime(const Item_func *func_item) {
...
+    if (this->array_schema->array_type() == TILEDB_DENSE)
+      DBUG_RETURN(func_item);
```
ihnorton, line 1311: will this give a useful error message?

mytile/ha_mytile.cc
```diff
@@ -1286,26 +1295,111 @@ const COND *tile::mytile::cond_push_func_datetime(const Item_func *func_item) {
...
+    int ret = set_range_from_item_datetime(
+        ha_thd(), dynamic_cast<Item_cache_datetime *>(args[1]),
+        dynamic_cast<Item_cache_datetime *>(args[1]), cmp_type, range,
+        datatype);
+
+    if (ret)
+      DBUG_RETURN(func_item);
+
+    // If this is an attribute add it to the query condition
+    if (this->query_condition == nullptr) {
```
ihnorton, line 1392: Do we need to check `use_query_condition` here like below?

ihnorton: nevermind, I see this is handled above at the case entry already, and this clause is different than the ones that check `use_query_condition` before creating the QC.

ihnorton (approved):
Minor ?s but LGTM.

Not necessarily an issue for right now, but as a general question: is there any way we can assert that the query condition is actually applied? Or put another way: would the SQL test clauses still pass if the conditions are not pushed down to libtiledb?

mytile/ha_mytile.cc
ihnorton: (or does it just do the filtering in MariaDB instead?)

mytile/ha_mytile.cc
Shelnutt2: I don't really know, I was getting errors from https://github.com/TileDB-Inc/TileDB/blob/ea3f4c8aac36d66db4ff5167f5a8e2627a4e6601/tiledb/sm/query/reader.cc#L2380

mytile/ha_mytile.cc
Shelnutt2: This makes MariaDB do the filtering instead.

PR #173, opened 2021-09-20
Fix index handling in 10.5

Shelnutt2:
This PR has a few fixes, first we set the dimension order to be the same as the index orders. Next we properly handle MRR keys to only increment the last_key part. That is for MariaDB when using non-inclusive ranges, only the last part of the key is expected to be incremented.

Lastly we handle the key_length not being the same as the TileDB datatype.

### DIFF
--- a/mytile/mytile-range.h
+++ b/mytile/mytile-range.h
@@ -686,10 +686,13 @@ void update_range_from_key_for_super_range(std::shared_ptr<tile::range> &range,
   // TileDB ranges are inclusive
   case Item_func::GT_FUNC: {
     range->operation_type = Item_func::GE_FUNC;
-    if (std::is_floating_point<T>()) {
-      *key_value = std::nextafter(*key_value, std::numeric_limits<T>::max());
-    } else if (!std::is_same<T, char>()) {
-      *key_value += 1;
+    // Only increment last key part
+    if (last_key_part) {
+      if (std::is_floating_point<T>()) {
+        *key_value = std::nextafter(*key_value, std::numeric_limits<T>::max());
+      } else if (!std::is_same<T, char>()) {
+        *key_value += 1;
+      }
     }
 
     // If the lower is null, set it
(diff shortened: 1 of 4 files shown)


mytile/mytile-range.h
```diff
@@ -753,10 +756,14 @@ void update_range_from_key_for_super_range(std::shared_ptr<tile::range> &range,
...
     range->operation_type = Item_func::LE_FUNC;
-    if (std::is_floating_point<T>()) {
-      *key_value = std::nextafter(*key_value, std::numeric_limits<T>::min());
-    } else if (!std::is_same<T, char>()) {
-      *key_value -= 1;
+
+    // Only decrement last key part
+    if (last_key_part) {
+      if (std::is_floating_point<T>()) {
+        *key_value = std::nextafter(*key_value, std::numeric_limits<T>::min());
```
ihnorton, line 763: Should the direction here be conditional on which side of the range we are looking at?

Shelnutt2: You mean if its upper or lower of TIleDB Range?

ihnorton: Yes

ihnorton: I mean that we are going toward `std::numeric_limits<T>::min()` but IIUC that assumes this is the upper range

ihnorton: Oh, this is the last key so it has to be the upper by construction? I'm probably a little confused because some of the internal logic seems to imply that the keys can be out of order, but might have misunderstood (x-ref https://github.com/TileDB-Inc/TileDB-MariaDB/blob/4f2f1c26b0dc74570a87d769b74d47fb33124bf5/mytile/ha_mytile.cc#L2716)

ihnorton: Resolved after discussion.

antalakas (approved):

PR #184, opened 2022-02-25
Update for TileDB 2.6.3

Shelnutt2:
Update for TileDB 2.6.3


mytile/ha_mytile.cc
```diff
@@ -3347,7 +3347,7 @@ mysql_declare_plugin(mytile){
     PLUGIN_LICENSE_PROPRIETARY, /* the plugin license (PLUGIN_LICENSE_XXX) */
     mytile_init_func,           /* Plugin Init */
     NULL,                       /* Plugin Deinit */
-    0x0121,                     /* version number (0.12.1) */
```
robertbindar, line 3350: Update comments on the right to 0.12.5

robertbindar (approved):
Changes seem consistent with past updates. You missed to update some comments in ha_mytile, but apart from that, all good from my side.

mytile/ha_mytile.cc
Shelnutt2: Good catch!

PR #192, opened 2022-06-14
Disable MRR by default

Shelnutt2:
RR is most useful for joins and bulk key access (BKA). MariaDB however also attempts to use MRR if implemented for quick selects. This introduces a new sysvar `mytile_mrr_support` which toggles whether MRR is enabled or if its treated as disable and unavailable.

[sc-18106]


Shelnutt2:
This solves a user issue, so I'm going to merge and release it. I have a followup story to enable MRR automatically if the user requests it for a join operation.

PR #204, opened 2022-11-22
Add nightly build with GitHub Actions

jdblischak:
**Summary:** This PR adds a GitHub Actions workflow to build and test the latest TileDB-MariaDB with the latest libtiledb every night

Notes:

* Translated from the existing Azure pipeline. I was able to simplify by removing Azure-specific steps, and also the nightly build doesn't need to test `SUPERBUILD=OFF`
* Did my best to remove any GitHub Actions specific functionality. This makes it possible to run the entire pipeline locally, eg
    ```sh
    bash scripts/nightly/checkout.sh
    bash scripts/nightly/setup-ubuntu.sh
    bash scripts/nightly/build-libtiledb.sh
    bash scripts/nightly/build-tiledb-mariadb.sh
    ```

Plans for future improvements:

* Currently the CI scripts download and build the latest libtiledb from source. The next step is to modify `CMakeLists.txt` to have the option to automatically build the latest libtiledb from source
* Automatically open an Issue for a failed build

### DIFF
--- a/scripts/nightly/build-libtiledb.sh
+++ b/scripts/nightly/build-libtiledb.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -eux
+
+# Build libtiledb from source
+
+cd libtiledb
+mkdir -p build
+cd build
+cmake \
+  -DTILEDB_VERBOSE=ON \
+  -DTILEDB_S3=ON \
+  -DTILEDB_SERIALIZATION=ON \
+  -DCMAKE_BUILD_TYPE=Debug \
+  ..
+make -j2
+sudo make -C tiledb install
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,0 +1,38 @@
+name: nightly
+on:
+  push:
+    paths:
+      - '.github/workflows/nightly.yaml'
+      - 'scripts/nightly/**'
+  schedule:
+    - cron: "0 2 * * *" # Every night at 2 AM UTC (9 PM EST; 10 PM EDT)
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    name: nightly-${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["macos-12", "ubuntu-20.04"]
+    env:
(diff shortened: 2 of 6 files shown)


jdblischak:
Since this PR isn't going to build a brand new GitHub Actions workflow, you can see the build logs in my fork:

https://github.com/jdblischak/TileDB-MariaDB/actions/runs/3524910709

scripts/nightly/build-libtiledb.sh
```diff
@@ -0,0 +1,16 @@
...
+mkdir -p build
+cd build
+cmake \
+  -DTILEDB_VERBOSE=ON \
+  -DTILEDB_S3=ON \
+  -DTILEDB_SERIALIZATION=ON \
+  -DCMAKE_BUILD_TYPE=Debug \
+  ..
+make -j2
+sudo make -C tiledb install
```
Shelnutt2, line 16: Do we need a `sudo ldconfig` for linux?

Shelnutt2 (approved):
One comment, otherwise lgtm! Great work!!

scripts/nightly/build-libtiledb.sh
jdblischak: I don't know for sure. At the very least it isn't required for a successful build.

* I update `LD_LIBRARY_PATH` in `scripts/nightly/build-tiledb-mariadb.sh`. This is the same strategy as is used in the Azure pipeline. However, I see in the Docker files that they instead run `ldconfig`. Since I can't run `ldconfig` on macOS, I'd prefer to keep it as is (ie set `LD_LIBRARY_PATH` on Ubuntu and `DYLD_LIBRARY_PATH` on macOS)
* As far as I can tell, MyTile is able to find and use the locally compiled libtiledb, eg [here](https://github.com/jdblischak/TileDB-MariaDB/actions/runs/3524910709/jobs/5910944210#step:7:895)

Shelnutt2: Thanks for the details, I'm fine with the env vars. Was more asking since we use it in docker. Docker though are a little more generic and we assume users might inherit from our containers and thus potentially change env variables.

jdblischak:
The new nightly builds passed post-merge :-)

https://github.com/TileDB-Inc/TileDB-MariaDB/actions/runs/3526133624/jobs/5913698988

Going forward, they should only run nightly, unless the YAML or CI scripts are edited.

PR #215, opened 2023-01-03
Remove macOS installation of libjudy

ihnorton:
libjudy is required by MariaDB's optional OQGraph plugin, which we do not need for nightly testing. The homebrew package is unmaintained and regularly fails to install.

We'll still have coverage on linux.

Fixes https://github.com/TileDB-Inc/TileDB-MariaDB/issues/214


ihnorton:
cc @jdblischak

PR #228, opened 2023-01-18
Update to TileDB 2.14.0

Shelnutt2:
Update to TileDB 2.14.0

PR #229, opened 2023-01-18
Support UTF-8 string condition pushdown

Shelnutt2:
Support UTF-8 string condition pushdown. This depends on #228 . Unit tests are not yet added, @DimitrisStaratzis will do it in a followup as we need to enable easier creation of utf-8 datatype attributes.


jdblischak:
I was surprised that this PR triggered the nightly builds. However, I think I figured it out. When you rebased your branch today to include my recent commit that updated the nightly builds, and then pushed to an internal branch in this repo, that triggered GitHub Actions via a `push` trigger. That also explains why it wasn't triggered when the PR was first opened (as expected).

PR #234, opened 2023-01-28
remove usage of deprecated array constructors with encryption

DimitrisStaratzis:
Ran a manual nightly build here: https://github.com/TileDB-Inc/TileDB-MariaDB/actions/runs/4031658316


Shelnutt2 (changes requested):
We need to use the new constructors that support time traveling.

PR #235, opened 2023-02-02
adding support for OR pushdown

DimitrisStaratzis:


mytile/ha_mytile.cc
```diff
@@ -1282,32 +1283,69 @@ void tile::mytile::dealloc_buffers() {
...
+  // Depending on the condition type (OR, AND), we create a combination of
+  // conditions. Then we combine this combined Query Condition with the "master"
+  // Query Condition with an AND. In the end, the "master" Query Condition
+  // contains all the sub conditions.
   for (uint32_t i = 0; i < arglist->elements; i++) {
     if ((subitem = li++)) {
       // COND_ITEMs
-      cond_push_local(dynamic_cast<const COND *>(subitem));
+      cond_push_local(dynamic_cast<const COND *>(subitem), queryCondition);
+      // Dimensions do not support QCs, hence the queryCondition ptr
```
Shelnutt2, line 1313: Actually we do allow dimensions to have conditions now. Its still best to use a range or convert to a range where we can. However in some complex scenarios it makes sense. I.e `SELECT * FROM array WHERE (dim1 = 0 and attr1 = 1) OR (dim1 = 2 and attr1 = 3)`. In that case the idea thin is to grab `dim1` and set the range to `[0, 2]` but also apply the query condition to get the finer filtering.

mytile/ha_mytile.h
```diff
@@ -306,7 +309,8 @@ class mytile : public handler {
    * @param cond condition being pushed
    * @return condition stack which could not be pushed
    */
-  const COND *cond_push_local(const COND *cond);
+  const COND *cond_push_local(const COND *cond,
```
Shelnutt2, line 312: Please add the `QueryCondition` parameter to the doc strings

mytile/ha_mytile.h
```diff
@@ -266,14 +266,17 @@ class mytile : public handler {
...
+  const COND *cond_push_func(const Item_func *func_item,
+                             std::shared_ptr<tiledb::QueryCondition> &pPtr);
 
   /**
    * Handle datetiem function pushdowns
    * @param func_item
    * @return
    */
-  const COND *cond_push_func_datetime(const Item_func *func_item);
+  const COND *
```
Shelnutt2, line 277: Please add the `QueryCondition` parameter to the doc strings

mytile/ha_mytile.h
```diff
@@ -266,14 +266,17 @@ class mytile : public handler {
    * @param func_item
    * @return
    */
-  const COND *cond_push_func(const Item_func *func_item);
+  const COND *cond_push_func(const Item_func *func_item,
```
Shelnutt2, line 269: Please add the `QueryCondition` parameter to the doc strings

Shelnutt2 (approved):
Looking good! A few minor nits. The largest one about supporting dimension query condition, I think this should be a separate followup PR.

PR #245, opened 2023-04-27
update to TileDB 2.15.2

DimitrisStaratzis:


mytile/ha_mytile.cc
```diff
@@ -3362,7 +3362,7 @@ mysql_declare_plugin(mytile){
     PLUGIN_LICENSE_PROPRIETARY, /* the plugin license (PLUGIN_LICENSE_XXX) */
     mytile_init_func,           /* Plugin Init */
     NULL,                       /* Plugin Deinit */
-    0x0222,                     /* version number (0.22.2) */
```
Shelnutt2, line 3365: Do we need to bump the version again? We didn't release a 0.22.2 yet right?

Shelnutt2 (approved):
One comment, then LGTM

mytile/ha_mytile.cc
DimitrisStaratzis: no we don't you are right!

PR #246, opened 2023-05-03
fixing issue 242

DimitrisStaratzis:
This PR fixes issue #242. The problem was that libtiledb could not be found by ```find_package(TileDB QUIET)```. The issue was resolved by adding the ```--prefix``` cmake flag.

### DIFF
--- a/scripts/nightly/build-libtiledb.sh
+++ b/scripts/nightly/build-libtiledb.sh
@@ -7,6 +7,7 @@ cd libtiledb
 mkdir -p build
 cd build
 cmake \
+  -DCMAKE_INSTALL_PREFIX=/usr/local/ \
   -DTILEDB_VERBOSE=ON \
   -DTILEDB_S3=ON \
   -DTILEDB_SERIALIZATION=ON \


Shelnutt2 (changes requested):
We should not use the bootstrap script. Please stick with cmake and add the cmake prefix option.

Shelnutt2 (approved):

DimitrisStaratzis:
> We should not use the bootstrap script. Please stick with cmake and add the cmake prefix option.

Done! Is there a reason not to use the bootstrap script?

Shelnutt2:
> > We should not use the bootstrap script. Please stick with cmake and add the cmake prefix option.
> 
> Done! Is there a reason not to use the bootstrap script?

Yes, the bootstrap script just wraps the exact same basic calls but introduced yet another layer into the build fix. There are more cmake options than what the bootstrap script exposes, so best to keep with cmake so we have all the options open.

DimitrisStaratzis:
> I created the nightly build scripts based off of the existing CI scripts. Should the call to `cmake` in [`scripts/ci_install_tiledb_and_run_tests.sh`](https://github.com/TileDB-Inc/TileDB-MariaDB/blob/b2f3b9b324ed118abd147aaba6af3bcf377a182a/scripts/ci_install_tiledb_and_run_tests.sh) also be updated to include `-DCMAKE_INSTALL_PREFIX=/usr/local/`?
> 
> https://github.com/TileDB-Inc/TileDB-MariaDB/blob/b2f3b9b324ed118abd147aaba6af3bcf377a182a/scripts/ci_install_tiledb_and_run_tests.sh#L30

The status messages below indicate that libtiledb is found as expected:

https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=31922&view=logs&j=f4ee9066-1821-5d26-c193-d558ffef2269&t=db128dfa-18a1-5b21-9fe0-ac1172e703e3&l=1385

https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=31922&view=logs&j=12357780-754c-5390-0d30-b98240e538d2&t=39c55a51-8f32-5817-6643-3e745179c7e2&l=1296

If that wasn't the case then ```TileDB not found for MyTile. Building as dependency from source``` would appear. Building from source in MacOS seems to be the underlying issue of this issue.

PR #250, opened 2023-06-02
fix string dim slicing bug

DimitrisStaratzis:
Read shortcut story for more details about the bug. 

I reproduced it locally by downloading the TileDB array stored in s3 https://cloud.tiledb.com/arrays/details/TileDB-Inc/504e9cc6-e8c4-4609-a5a4-1d243a9a7d1b/overview. 

The bug was happening when we were reading an array with dims of mixed type. (String and Numeric). More specifically,  when filtering on the String dim only, MyTile was failing to verify that the other dim matched the MariaDB key. This checking algorithm is part of MyTile  and was modified to avoid the bug. 

I am not including any new tests in this PR since the only array (as far as I know) that had this bug is 600MB. In smaller arrays the bug was not present. (Multiple read batches are required). 

A full reproduction can be found here: https://cloud.tiledb.com/notebooks/details/dstara/68c1c8c9-9209-4c34-bb1e-dabfc3ce9869/overview

Other than the bug fix, this PR also includes a minor code change which, after debugging, I found out that was adding the same range for each dimension multiple times and by multiple times I mean  ```X```(read-batches)

### DIFF
--- a/mytile/ha_mytile.cc
+++ b/mytile/ha_mytile.cc
@@ -2539,8 +2539,8 @@ void tile::mytile::open_array_for_reads(THD *thd) {
 
   // Set layout
   this->query->set_layout(query_layout);
-  // If the layout is unordered by the the array type is dense we will set it to
-  // the tile order instead Currently multi-range queries do not support dense
+  // If the layout is unordered but the array type is dense we will set it to
+  // the tile order instead. Currently, multi-range queries do not support dense
   // unordered reads
   if (this->array_schema->array_type() == tiledb_array_type_t::TILEDB_DENSE &&
       query_layout == tiledb_layout_t::TILEDB_UNORDERED) {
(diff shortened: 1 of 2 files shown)


mytile/ha_mytile.cc
```diff
@@ -3084,13 +3084,13 @@ int tile::mytile::set_pushdowns_for_key(const uchar *key, uint key_len,
...
-    for (uint64_t i = 0; i < this->ndim; i++) {
-      auto range = ranges_from_keys[i];
-      if (range.get() != nullptr) {
-        this->pushdown_ranges[i].push_back(range);
+      // Copy shared pointer to main range pushdown
+      for (uint64_t i = 0; i < this->ndim; i++) {
+        auto range = ranges_from_keys[i];
+        if (range.get() != nullptr) {
+          this->pushdown_ranges[i].push_back(range);
+        }
```
DimitrisStaratzis, line 3093: This is the minor change other than the bug fix

ihnorton:
> I am not including any new tests in this PR since the only array (as far as I know) that had this bug is 600MB. In smaller arrays the bug was not present. (Multiple read batches are required).

Can we change the read batch size in order to trigger it on a small array?

DimitrisStaratzis:
> > I am not including any new tests in this PR since the only array (as far as I know) that had this bug is 600MB. In smaller arrays the bug was not present. (Multiple read batches are required).
> 
> Can we change the read batch size in order to trigger it on a small array?

I am unable to reproduce it in a small array. I tried changing the ```mytile_read_buffer_size``` and many other options but apparently it has to do with how the MariaDB optimizer initiates the query. That is the reason I used the TileDB-Cloud array in the first place.

PR #253, opened 2023-06-08
Add a pushdown for geometry types

perrygeo:
Changes:
- Adds a `cond_push_func_spatial` function to handle  `ST_Intersection` and similar spatial functions, which adds a range pushdown to `_X` and `_Y` dimensions. See [paper doc](https://www.dropbox.com/scl/fi/pd36ldxrh7lz1p08km6so/MariaDB-Spatial-SQL.paper?dl=0&rlkey=nvkvcvaqqmu2zgy0drz2x6czs) for more details.
- Map the MySQL Blob types to TILEDB_BLOB. This enables creation of blob attributes with CREATE SQL statements (previously resulting in char).

### DIFF
--- a/mysql-test/mytile/r/spatial_index.result
+++ b/mysql-test/mytile/r/spatial_index.result
@@ -1,0 +1,48 @@
+# SPATIAL INDEX PUSHDOWN
+# Create a test array 
+create table t1 (
+_X           double DIMENSION=1 lower_bound="0" upper_bound="100" tile_extent="10",
+_Y           double DIMENSION=1 lower_bound="0" upper_bound="100" tile_extent="10",
+name         varchar(255) filters="GZIP=-1",
+wkb_geometry blob NOT NULL
+)
+engine=MyTile
+array_type='SPARSE'
+;
+# Insert 2 polygon geometries
+set @g1 = GeometryFromText('POLYGON((10.0 10.0, 20.0 10.0, 20.0 20.0, 10.0 20.0, 10.0 10.0))');
+set @g2 = GeometryFromText('POLYGON((84.0 84.0, 94.0 84.0, 94.0 94.0, 84.0 94.0, 84.0 84.0))');
+set @bbox1 = ST_Centroid(ST_Envelope(@g1));
+set @bbox2 = ST_Centroid(ST_Envelope(@g2));
+insert into t1 (_X, _Y, name, wkb_geometry) values
+(ST_X(@bbox1), ST_Y(@bbox1), "building1", aswkb(@g1)),
+(ST_X(@bbox2), ST_Y(@bbox2), "building2", aswkb(@g2))
+;
+# Test that only one intersects a search radius
+set @point_text = 'POINT(90.0 90.0)';
+set @buffer_distance = 10;
+select name, ST_Area(GeometryFromWkb(wkb_geometry))
+from t1
+where ST_Intersects(
+ST_Buffer(GeometryFromText(@point_text), @buffer_distance),
+GeometryFromWkb(wkb_geometry)
+);
+name	ST_Area(GeometryFromWkb(wkb_geometry))
+building2	100
+Warnings:
+Warning	1292	Truncated incorrect DOUBLE value: '\x00\x00\x00\x00\x01\x03\x00\x00\x00\x01\x00\x00\x00\x7F\x00\x00\x00\x00\x00\x00\x00\x00\x80V@\x00\x00\x00\x00\x00\x00T@\xB2\...'
+# Test again, but without the spatial index
+# (results should be identical, but without pushdown to _X _Y)
+set @mytile_enable_pushdown = false;
+select name, ST_Area(GeometryFromWkb(wkb_geometry))
+from t1
+where ST_Intersects(
+ST_Buffer(GeometryFromText(@point_text), @buffer_distance),
+GeometryFromWkb(wkb_geometry)
+);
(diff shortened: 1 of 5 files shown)


mytile/ha_mytile.cc
```diff
@@ -1616,6 +1618,204 @@ const COND *tile::mytile::cond_push_func_datetime(
...
+        geom->get_mbr(&mbr, &c_end);
+        x1 = mbr.xmin;
+        y1 = mbr.ymin;
+        x2 = mbr.xmax;
+        y2 = mbr.ymax;
+      } else {
+        std::cout << "ERROR! Geometry is not valid, using dummy mbr" << std::endl;
+      }
+    }
+    // ----------------- END SKETCHY SECTION --------------------- //
```
perrygeo, line 1740: As the comment says, this is the sketchy part :-) 

I need to eval the `Item_cache` and somehow cast it to the native `Geometry` so that I have access to the spatial methods like `get_mbr`. This is my current attempt with the `val_native_with_conversion_result` method which obviously doesn't work. I'm struggling to find good examples or documentation about this. Does anyone see an obvious solution.

DimitrisStaratzis: Fixed!

mytile/ha_mytile.cc
```diff
@@ -1616,6 +1618,187 @@ const COND *tile::mytile::cond_push_func_datetime(
...
+      THD *thd = ha_thd();
+
+      // X dimension
+      {
+        std::shared_ptr<range> range = std::make_shared<tile::range>(tile::range{
+            std::unique_ptr<void, decltype(&std::free)>(nullptr, &std::free),
+            std::unique_ptr<void, decltype(&std::free)>(nullptr, &std::free),
+            Item_func::BETWEEN, tiledb_datatype_t::TILEDB_ANY, 0, 0});
+
+        Item_float *lower = new (thd->mem_root) Item_float(thd, x1, 0);
```
Shelnutt2, line 1766: Do lower/upper need to be free'ed?

perrygeo: I'm not sure. Freeing lower/upper leads to a segfault, which I believe is due to `tile::range` accessing them later, i.e. the range owns the items. I'll continue looking into it.

Shelnutt2: If the ownership is passes to `range` that is fine. I just saw the `new` call with out `delete` so wanted to check.

mysql-test/mytile/r/spatial_index.result
```diff
@@ -0,0 +1,48 @@
...
+set @bbox1 = ST_Centroid(ST_Envelope(@g1));
+set @bbox2 = ST_Centroid(ST_Envelope(@g2));
+insert into t1 (_X, _Y, name, wkb_geometry) values
+(ST_X(@bbox1), ST_Y(@bbox1), "building1", aswkb(@g1)),
+(ST_X(@bbox2), ST_Y(@bbox2), "building2", aswkb(@g2))
+;
+# Test that only one intersects a search radius
+set @point_text = 'POINT(90.0 90.0)';
+set @buffer_distance = 10;
+select name, ST_Area(GeometryFromWkb(wkb_geometry))
```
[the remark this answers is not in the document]
perrygeo: I got rid of the ST_AREA call but I'm still getting that warning. At least it's consistent, but I do worry that MariaDB updates might change the output and cause the test to fail in the future. I don't see it in the client or server logs when I run it manually, only during the integration tests.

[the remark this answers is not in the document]
DimitrisStaratzis: I will work on it tomorrow

mytile/ha_mytile.cc
```diff
@@ -1616,6 +1618,183 @@ const COND *tile::mytile::cond_push_func_datetime(
...
+  int wkb_arg = -1;
+  int aoi_arg = -1;
+  for (uint i = 0; i < func_item->argument_count(); i++) {
+    switch (args[i]->type()) {
+    case Item::FIELD_ITEM: {
+      Item_field *column_field = dynamic_cast<Item_field *>(args[i]);
+      if (column_field != nullptr) {
+        if (column_field->field_name.str == geometry_column) {
+          // The bare field name references a blob, should be a type error
+          DBUG_RETURN(nullptr);
```
DimitrisStaratzis, line 1652: Throw an error maybe?

mytile/ha_mytile.cc
```diff
@@ -1616,6 +1618,183 @@ const COND *tile::mytile::cond_push_func_datetime(
...
+    if (aoi->has_value()) {
+      String arg_val;
+      String *swkb = aoi->val_str(&arg_val);
+      Geometry_buffer buffer;
+      Geometry *geom = NULL;
+
+      geom = Geometry::construct(&buffer, swkb->ptr(), swkb->length());
+
+      // Calculate minimum bounding rectangle of geometry
+      if (geom != nullptr) {
```
DimitrisStaratzis, line 1712: throw an error if geom is nullptr which means that it was not given correctly by user.

mysql-test/mytile/t/spatial_index.test
DimitrisStaratzis, line 1: Could we add more tests for all the supported functions?

```SP_TOUCHES_FUNC```, ```SP_CROSSES_FUNC```, ...

DimitrisStaratzis (approved):
I have checked that the pushdown works by comparing the number of records TileDB returns. I have only left minor comments, great work!

PR #255, opened 2023-06-13
add support for ALTER TABLE ADD/DROP

DimitrisStaratzis:
This PR adds support for the ```ALTER TABLE ADD/DROP``` operations in MyTile by using the ```SchemaEvolution``` API. All other requests will return an error message.

Users can now drop one or more columns by name:
```ALTER TABLE `table` DROP `a3`, DROP `a4`;```

Or add new attributes and configure them the same way they do when they use ```CREATE_TABLE```
```ALTER TABLE `table` ADD COLUMN (`a10` int filters='GZIP=6');```

[sc-30285]

### DIFF
--- a/mytile/ha_mytile.cc
+++ b/mytile/ha_mytile.cc
@@ -57,6 +57,7 @@
 #include <vector>
 #include <unordered_map>
 #include <key.h> // key_copy, key_unpack, key_cmp_if_same, key_cmp
+#include "tiledb/array_schema_evolution.h"
 
 // Handler for mytile engine
 handlerton *mytile_hton;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: mytile/ha_mytile.cc [hunk 1]
```diff
@@ -337,6 +338,133 @@ int tile::mytile::close(void) {
...
+    // drop columns
+    for (const std::string& col : columns_to_be_dropped) {
+      evolution.drop_attribute(col.c_str());
+    }
+    // evolve array
+    evolution.array_evolve(this->uri);
+
+    DBUG_RETURN(false); //success
+
+    // if the evolution is a DROP column
```
Shelnutt2, line 430: Comment should be `ADD`?

Shelnutt2 (approved):
Minor nit, otherwise great work!

mytile/ha_mytile.cc
DimitrisStaratzis: Thanks! Done

PR #274, opened 2023-09-14
Update to TileDB 2.17.0

Shelnutt2:
Update to TileDB 2.17.0


DimitrisStaratzis (approved):
Approving if CI passes. After looking at the new enumerated datatypes feature it is not obvious to me whether changes are needed in the read-path. I mean if we need to make changes to be able to read arrays with enums. Changes are certainly needed in the write-path. These are all coming in my next PR which will be pushed ASAP.

PR #275, opened 2023-09-21
Dstara/add support for enum types

DimitrisStaratzis:
# Description
This PR adds full support for the TileDB Enum types to Mytile. 

## Read path
Arrays created with any TIleDB API that have enums will work seamlessly with mytile. Mytile will recognize the enums and display them accordingly:

Example:

```DESCRIBE tiledb_enum;```
| column_name | COLUMN_TYPE               | is_nullable | column_default |
|-------------|---------------------------|-------------|----------------|
| a0          | enum('ee','ff','gg','hh') | NO          | NULL           |
| a1          | enum('aa','bb','cc','dd') | NO          | NULL           |
| a2          | enum('ii','jj','kk','ll') | NO          | NULL           |
| a3          | int(11)                   | NO          | NULL           |

In cases where TIleDB uses an enum with numerals, mytile will ignore the enum. This is the case for the ```a3``` attribute above. MariaDB does not support numerals in enums. We could convert the numerals from TileDB to strings but that would make users unable to perform numerical operations in these attributes. (Discussed with @ihnorton about that)

## Write path
Whenever using mytile to create an array with an enum, a TileDB array with a TileDB enum will be created. I attach such an array so you can review its contents and possibly check against other APIs. I tested it with the Java API.

Array with enum created by Mytile:
[tiledb_with_enum.zip](https://github.com/TileDB-Inc/TileDB-MariaDB/files/12689984/tiledb_with_enum.zip)

## Query Conditions
Filtering with Query Conditions is not currently enabled for enum types. I tried to implement it but even though the filtering works in the TileDB level and I can see the records returned, I have yet to figure out how to set the MariaDB fields. I will work on that in an upcoming PR.

### CODE REVIEW

File: mytile/ha_mytile.cc [hunk 1]
```diff
@@ -1549,6 +1569,7 @@ const COND *tile::mytile::cond_push_func_datetime(
 
   switch (func_item->functype()) {
   case Item_func::ISNULL_FUNC: {
+    DBUG_RETURN(func_item);
```
ihnorton, line 1572: This duplicates the DBUG_RETURN below?

mytile/ha_mytile.cc
```diff
@@ -1963,14 +1986,13 @@ tile::mytile::cond_push_func(const Item_func *func_item,
...
 
-    // Dense arrays do not support query conditions
-    if (this->array_schema->array_type() == TILEDB_DENSE)
-      DBUG_RETURN(func_item);
-
     auto attr = this->array_schema->attribute(column_field->field_name.str);
     datatype = attr.type();
     nullable = attr.nullable();
+    auto enmr_name = tiledb::AttributeExperimental::get_enumeration_name(ctx, attr);
+    is_enum = enmr_name.has_value();
```
ihnorton, line 1993: This doesn't seem like the right way to check if an attribute is an enum

mytile/mytile-discovery.cc
```diff
@@ -285,14 +285,42 @@ int tile::discover_array(THD *thd, TABLE_SHARE *ts, HA_CREATE_INFO *info) {
...
+      bool is_enum = enmr_name.has_value();
+
       auto mysql_type =
           TileDBTypeToMysqlType(attribute.type(), attribute.cell_size() > 1, attribute.cell_val_num());
-      if (mysql_type == MYSQL_TYPE_VARCHAR) {
-        sql_string << "TEXT";
+
+      if (is_enum) {
+        // if the attribute has an enum
+        auto array = tiledb::Array(ctx, array_uri_timestamp.first, TILEDB_READ,
```
ihnorton, line 296: Is it possible to cache this so that we don't have to reopen elsewhere? Opening is relatively expensive for remote arrays.

mytile/ha_mytile.cc
davisp: Whoops, that's my bad. That is the correct logic, but I should have provided a `bool AttributeExperimental::has_enumeration(ctx, attr)` wrapper.

mytile/mytile-discovery.cc
DimitrisStaratzis: Yes but we will have to make a lot of changes so I think it is better if we wait for a getEnumeration() that takes array schema as a param.

mytile/ha_mytile.cc
```diff
@@ -1963,14 +1986,13 @@ tile::mytile::cond_push_func(const Item_func *func_item,
...
 
+    if (is_enum) DBUG_RETURN(func_item); //disable enum push down for now
```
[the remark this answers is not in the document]
DimitrisStaratzis: Done. https://app.shortcut.com/tiledb-inc/story/35299/add-support-for-enums-pushdown-with-qc

mytile/ha_mytile.cc
DimitrisStaratzis: No it disables pushdown for conditions where (value == null) because it was causing problems. I don't know how to construct a query condition to check equality with null. The one currently implemented is wrong and does not work correctly.

mytile/mytile-discovery.cc
DimitrisStaratzis: For now I have moved the array opening outside the for loop as we discussed.

mytile/mytile-discovery.cc
```diff
@@ -285,14 +285,42 @@ int tile::discover_array(THD *thd, TABLE_SHARE *ts, HA_CREATE_INFO *info) {
...
+                                   encryption_key.empty() ? TILEDB_NO_ENCRYPTION
+                                                          : TILEDB_AES_256_GCM,
+                                   encryption_key);
+        auto enmr = tiledb::ArrayExperimental::get_enumeration(
```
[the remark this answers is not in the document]
DimitrisStaratzis: I opened a story to track this down. 
https://app.shortcut.com/tiledb-inc/story/35300/replace-arrayexperimental-get-enumeration-ctx-array-enmr-name-value-with-arrayexperimental-get-enumeration-ctx-array-schema

mytile/mytile-discovery.cc
```diff
@@ -316,21 +344,23 @@ int tile::discover_array(THD *thd, TABLE_SHARE *ts, HA_CREATE_INFO *info) {
...
-                                         default_value_size) ||
-            !tile::is_string_datatype(attribute.type())) {
-          auto default_value_str = TileDBTypeValueToString(
-              attribute.type(), default_value, default_value_size);
-
-          if (!default_value_str.empty())
-            sql_string << " DEFAULT " << default_value_str;
-        }
+      if (!is_enum){
+          if (valid == 0)
```
ihnorton, line 348: nit: braces

ihnorton (approved):

PR #279, opened 2023-10-02
updating ubuntu image to ```ubuntu-latest``` for nightly builds

DimitrisStaratzis:
This compilation error was caused because we were using gcc 9. I have updated the nightly script to use ```ubuntu-22.04``` which uses gcc 11. I think it's better than manually updating the gcc in the older image.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["macos-13", "ubuntu-20.04"]
+        os: ["macos-13", "ubuntu-22.04"]
     env:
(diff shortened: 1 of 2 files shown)


ihnorton (approved):
This is OK, but just be aware that we may hit issues next time the `-latest` image changes, whereas 22.04 should be stable for quite a while (18.04 was removed in 2022), and we can pick when to update based on compiler needs.

DimitrisStaratzis:
> This is OK, but just be aware that we may hit issues next time the `-latest` image changes, whereas 22.04 should be stable for quite a while (18.04 was removed in 2022), and we can pick when to update based on compiler needs. Up to you which one to choose.

Yes you are right, changing to 22.04. Will merge upon successful CI.

PR #289, opened 2023-11-21
update to TileDB-2.18.0

DimitrisStaratzis:


CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ if (CMAKE_VERSION VERSION_LESS "3.1")
         set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++17")
     endif ()
 else ()
-    set (CMAKE_CXX_STANDARD 17)
+    set (CMAKE_CXX_STANDARD 20)
```
DimitrisStaratzis, line 12: only questionable change of this PR

ihnorton: This should not be necessary for using the external C++ API... Were you getting build errors without it?

That said, we don't have any downstream users of MariaDB C++ as far as I know, so this should be fine.

ihnorton (approved):

CMakeLists.txt
DimitrisStaratzis: No I just thought to update it. There are no build errors when using CXX STARDARD 17

PR #292, opened 2023-12-07
Fix mysqld crash on Macos

robertbindar:
The exported status variables array should end with a null-like struct, the absence of that final null-like struct causes mysqld to crash on macos.
@Shelnutt2 I've mentioned this to you when I last debugged something in mytile, but back then I just worked around and moved on.


Shelnutt2 (approved):
Thanks for putting this in.

DimitrisStaratzis (approved):

PR #293, opened 2023-12-07
Set backup value of TILEDB_FORCE_ALL_DEPS to OFF in nightly build

jdblischak:
The CMake variable `TILEDB_FORCE_ALL_DEPS` should be set to `ON` or `OFF`. I accidentally had the backup value set to `NO`

This change will have no immediate effect since I set the env var `TILEDB_FORCE_ALL_DEPS` in the nightly workflow file:

https://github.com/TileDB-Inc/TileDB-MariaDB/blob/cacc61994e2b9b9a37387eea2bfc14220c375cc4/.github/workflows/nightly.yaml#L23


Shelnutt2 (approved):

jdblischak:
I also realized that the nightly workflow wasn't running on this PR, so I added a `pull_request` trigger

PR #298, opened 2023-12-12
Initial Azure to GH CI/CD changes

dudoslav:
Migrate CI/CD form Azure to Github Actions.

Two new workflows are implemented:

1. Stable
Called always, with steps nearly the same as the Nightly workflow with one exception, different REF_LIBTILEDB version. If this workflow passes AND is default branch OR tag ref then it also calls Docker workflow. See below.

2. Docker
Is a reusable workflow with matrix for each Dockerfile/repository. This workflow logs into dockerhub and subsequently build docker container image. Tagging logic is split depending on the event branch.

- If the branch is default branch then the container image created is tagged [REPO]:dev
- If the branch is a tag ref then the container image created is tagged [REPO]:latest and [REPO]:[TAG]

Example docker command for tag ref docker workflow:
```
/usr/bin/docker buildx build --build-arg MYTILE_VERSION=test_tag04 --file docker/Dockerfile-min --iidfile /home/runner/work/_temp/docker-actions-toolkit-8h4RdJ/iidfile --tag tiledb/tiledb-maria-min:latest --tag tiledb/tiledb-maria-min:test_tag04 --metadata-file /home/runner/work/_temp/docker-actions-toolkit-8h4RdJ/metadata-file .
```

Note that all pipelines in my forked repo will inevitably fail because Dockerfiles use hard coded original repo and are now aware that default branch can have '/' characters in it. This can be fied, but there is no need, since the default branch for the original repo is `master`.

Also you can see how the workflows are triggered here: https://github.com/dudoslav/TileDB-MariaDB/actions


dudoslav:
> Can we not remove the azure pipelines files to ensure that this works in master as intended and remove them in a follow-up PR?

Yes we can, but I'll make sure that Docker logic will not push the image twice (So I'll login but not push)

.github/workflows/stable.yaml
```diff
@@ -0,0 +1,38 @@
+name: stable
+
+# This workflow should run always
+on: [push]
```
DimitrisStaratzis, line 4: I think we should change that to run explicitly on pull requests, tags, pushes to master, and manually:

```
  workflow_dispatch:
  push:
    tags:
      - '*'
    branches:
      - 'master'
  pull_request:
```

dudoslav: Don't you want to run these tests on any kind of branch?

DimitrisStaratzis: No, let's not overload the runners with every push. Only in PRs. Also with ```workflow_dispatch``` we can run the CI manually when we want

dudoslav: Ok that makes sense, on it

DimitrisStaratzis (approved):
Great work!!! Thank you @dudoslav

PR #301, opened 2023-12-18
Add superbuild configuration into CI/CD stable matrix

dudoslav:
This PR adds superbuild variant into CI/CD matrix for Linux platform only, since the Mac OS superbuild does not work yet.


DimitrisStaratzis (approved):
Great! See only if we could add "superbuild" in the name.

dudoslav:
Yea you are right, I kind of thought that these github workflows will contain configuration in their name :/

dudoslav:
Done

PR #304, opened 2023-12-20
MRR queries overhaul

DimitrisStaratzis:
# Description of key changes

## Fix wrong mrr flag reset. 
This flag was being reset falsely after every batch read in MRR queries. This was resulting in MRR queries to not being MRR queries after the first batch. 

## Simplify ```index_read_scan()```

The previous implementation relied on the assumption that keys from MariaDB would arrive in a sorted order. However, @robertbindar discovered that this was not always the case. The new algorithm, on the other hand, does not assume that the keys are sorted but leverages the fact that they may be partially sorted. Specifically, the changes involve eliminating the index-- portion. The index now only increments forward. In the event that the keys are sorted, the algorithm takes advantage of this by resuming the index from where it left off each time, and if the index reaches the end of the records, it resets to zero. This modification essentially allows for a full cycle in case the keys arrive in orders like this: 1000, 1001, 1002, 1, 2, 3. The worst case scenario for this algorithm is to receive the keys in descending sorted order which is never the case. 

## Cache dimension names and domain for repetitive use in mrr queries. 
This was slowing down the execution massively. I found it by using the Apple-Instruments time profiler.  

## Results 
The query runtime has been dropped from an infinite loop to about 8 seconds. Initially the infinite loop was resolved by the new algorithm. This made the query complete in 40seconds. The domain caching then dropped the time to 8 seconds. (Tested locally in my laptop).

## Note
The join result in the new test is validated by TileDB-Trino and the user himself.

[sc-37142]


Shelnutt2:
> Enable UTF-8 pushdown

Please separate this out as a separate PR so we can merge it independently (i.e quicker).

mytile/ha_mytile.cc
```diff
@@ -3314,85 +3313,24 @@ int tile::mytile::index_read_scan(const uchar *key, uint key_len,
            (find_flag == ha_rkey_function::HA_READ_AFTER_KEY ||
             find_flag == ha_rkey_function::HA_READ_KEY_OR_NEXT))) {
         tileToFields(record_index, false, table);
-        this->record_index++;
+        found = true;
         this->records_read++;
+        this->record_index=0;
```
Shelnutt2, line 3318: Doesn't this mean we are doing a linear scan of the records for every single index next? That seems far from idea? Whats wrong with the previous logic of keeping the current position and finding if we should go forward or backwards in the results? If the keys come from MariaDB sorted this will result in a large performance hit I believe,

DimitrisStaratzis: Thanks for the comment! Nothing wrong with the optimization. I am testing a revised implementation of it. This is still WIP.

DimitrisStaratzis:
Update:
I have now added the optimization to take advantage of sorted keys but not depend on it. The triple join performance improved by about 15%. Some tests are still failing as you can see in the CI. Getting there!

mytile/ha_mytile.cc
```diff
@@ -453,15 +459,14 @@ enum_alter_inplace_result tile::mytile::check_if_supported_inplace_alter(
   DBUG_ASSERT(ha_alter_info != nullptr);
 
   // checking if ALTER operation is supported
-  if (ha_alter_info->handler_flags &
-      (ALTER_ADD_COLUMN | ALTER_DROP_COLUMN )) {
-    DBUG_RETURN(HA_ALTER_INPLACE_EXCLUSIVE_LOCK); //todo check lock type
+  if (ha_alter_info->handler_flags & (ALTER_ADD_COLUMN | ALTER_DROP_COLUMN)) {
```
Shelnutt2, line 462: There is a lot of unchanged lines that are just formatting changes. What formatter have you moved too? Can you do the formatting in a separate PR so its easier to track the real changes in this one?

DimitrisStaratzis: I used clang-format but I have forgotten to do it in previous PRs so that's why. I will remove these changes to make this PR cleaner before I open it for review. That's happening tomorrwo first thing

Shelnutt2: Formatting is good, but maybe we merge as a separate PR, then rebase this so its just a little clearer?

DimitrisStaratzis: Done!

DimitrisStaratzis: Will do a follow up PR for formatting

robertbindar (approved):
Great work @DimitrisStaratzis and sorry this took so long. I think this is a good fix for the the issue the user experienced.
To reiterate.
Step0) was the UTF8 condition pushdown fix
Step1) Get a naive version which scans through the keys and restarts if it completely missed the pivot point, but also assume keys may come sorted (avoid massive performance regression). This is implemented here.
Step2) The step1) above gets us a significant speedup, but it is far from ideal. We should file a followup for investigating if we can build a map for the keys we fetch during `build_mrr_ranges` and query that when we get a `index_read_scan` call from MariaDB for MRR queuries. This should bring us much much closer to what we can do with libtiledb.

Lastly, there is a comment left inline about filing a followup for adding a `string_view` version of dimension.name(), same for dimension, this is a good catch and a copy that can slip easily with anybody but has massive impact. (provided it's confirmed in optimized build).

Other than this, thanks, great work!.

mytile/ha_mytile.cc
```diff
@@ -3090,10 +3090,10 @@ int8_t tile::mytile::compare_key_to_dims(const uchar *key, uint key_len,
        key_part_index < key_info->user_defined_key_parts; key_part_index++) {
 
     const KEY_PART_INFO *key_part_info = &(key_info->key_part[key_part_index]);
-    tiledb::Dimension dimension = domain.dimension(key_part_index);
 
     for (auto &dim_buffer : this->buffers) {
-      if (dim_buffer == nullptr || dim_buffer->name != dimension.name()) {
+      if (dim_buffer == nullptr ||
```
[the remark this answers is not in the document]
DimitrisStaratzis: If I understand your comment correctly, then yes the builds were all in release mode when I found this performance issue. I have seen it in various places as well. I will open a shortcut story and discuss with @ihnorton.

PR #305, opened 2023-12-21
utf8 pushdown

DimitrisStaratzis:
Tests will follow


Shelnutt2 (changes requested):
Can you add a test + a test with an explain showing the pushdown?

DimitrisStaratzis:
> Can you add a test + a test with an explain showing the pushdown?

Done

Shelnutt2 (approved):
Thanks! Looks great!

PR #309, opened 2024-01-19
aggregates pushdown

DimitrisStaratzis:
# Description

This PR pushes the (SUM, AVG, MIN, MAX, COUNT) aggregates down to TileDB. 

I have added a new option:  ```enable_aggregate_pushdown``` (OFF by default).

# Implementation details

To achieve the aggregate pushdown I have implemented the ```mytile_group_by_handle```. This handler is responsible for opening the TileDB array, applying ranges, query conditions, aggregates and finally setting the temp fields that are created by mariadb to store the aggregates results. The ranges and query conditions are created in the mytile handler. I simply get them from there. 

# Performance experiments. 
The new pushdown offers a 250X performance improvement in some cases. These experiments have been conducted on an intel Mac. The array is a 4GB allele_count array downloaded from S3. Also, the results from the pushed down aggregations are validated.

| Query               | Aggr. pushdown On | Aggr. pushdown Off | Notes|
|---------------------|------------------|-------------------|-------------------|
| ```SELECT SUM(count) FROM allele``` | **0.8 seconds**              | 130 seconds             | Basic|
| ```SELECT AVG(count) FROM allele``` | **0.6 seconds**             |131 seconds             | Basic|
| ```SELECT MAX(count) FROM allele ```| **0.9 seconds**              | 116 seconds              | Basic|
| ```SELECT MIN(count) FROM allele``` | **0.7 seconds**              | 114 seconds             | Basic|
| ```SELECT SUM(count) FROM allele where alt = 'T'``` | **22 seconds**              | 80 seconds             | With QC|
| ```SELECT SUM(count) FROM allele where pos > 90000``` | **17 seconds**              | 147 seconds       | With dim slicing |
| ```SELECT MAX(alt) FROM allele``` | **0.8 seconds**              | 157seconds      | In String attribute |

More examples can be found in the new tests. 

# Supported operations

All from this link: https://docs.tiledb.com/main/background/internal-mechanics/aggregates

# Limitations

- Multi-valued attributes are not supported.
- Aggregate pushdown does not work when using multiple functions on a column. e.g. ```SELECT COALESCE(SUM(count), 0) from allele;```. Such queries revert back to MariaDB filtering.
- Aggregate pushdown does not work with ```GROUP BYs``` as they are not supported by TileDB yet. Such queries revert back to MariaDB filtering.
[sc-39455]
- Aggregations on Dimensions are supported on Dense arrays only. They are currently disabled due to a bug in core. (Easy to enable)


mytile/ha_mytile.cc
```diff
@@ -1053,6 +1053,154 @@ int tile::mytile::init_scan(THD *thd) {
...
+            *this->query, field);
+  } else {
+      DBUG_RETURN(1);
+  }
+  default_channel.apply_aggregate(field, operation);
+  log_debug(thd, "Applied aggregate %s for attribute %s", aggregate_str.c_str(), field.c_str());
+  DBUG_RETURN(0);
+}
+
+bool tile::mytile::has_aggregate(THD *thd, const std::string &field,
```
Shelnutt2, line 1105: Instead of passing `aggregate_str` by reference lets use a tuple return with a std::optional. Also I suggest instead of a string we use an enum or the tiledb (or mariadb) enum for the aggregates. This way we aren't having strings inline we are doing comparisons against.

mytile/ha_mytile.cc
```diff
@@ -1053,6 +1053,154 @@ int tile::mytile::init_scan(THD *thd) {
...
+        }
+        if (field == column_with_aggregate) {
+          switch (isp->sum_func()) {
+          case Item_sum::SUM_FUNC:
+            aggregate_str = "SUM";
+            DBUG_RETURN(true);
+
+          case Item_sum::AVG_FUNC:
+            aggregate_str = "AVG";
+            DBUG_RETURN(tile::sysvars::enable_avg_aggregate_pushdown(thd));
```
Shelnutt2, line 1141: Why is this different that the others?

mytile/ha_mytile.cc
```diff
@@ -1053,6 +1053,154 @@ int tile::mytile::init_scan(THD *thd) {
...
+    }
+  }
+  DBUG_RETURN(false);
+}
+
+bool tile::mytile::field_is_aggregation_compatible(const std::string &field) {
+    DBUG_ENTER("tile::mytile::field_is_aggregation_compatible");
+    if (this->array_schema->has_attribute(field)){
+        auto attr = this->array_schema->attribute(field);
+        DBUG_RETURN(attr.cell_val_num() == 1);
```
Shelnutt2, line 1165: @KiterLuc or @robertbindar  are var length fields supported in aggregates?

KiterLuc: For a list of supported input datatypes and what their respective output datatypes are, see https://docs.tiledb.com/main/background/internal-mechanics/aggregates. In summary, we support var size inputs for min/max, var size nullable inputs for null count. The only time we output a var size value is for min/max on strings.

DimitrisStaratzis: Hey, thanks for the comment! I know they are supported but mytile does not support them. Multi valued attributes are handled like BLOBs. So if mytile reads a multi valued attribute, it sets the type to BLOB and stores the raw bytes. Not sure how aggregates can work like this..

Shelnutt2: String/varchar should be supported.
count/count null should be supported for all types
max/min/avg should return not supported

mytile/ha_mytile.cc
DimitrisStaratzis: AVG and SUM require the user to enable them as they can have an effect in the types assigned to fields by mytile. Enabling them will promote all numeric types to double. Please see description for a detailed explanation of that.

mytile/ha_mytile.cc
DimitrisStaratzis: Done! I am now using the ```Item_sum::sumfunctype``` enum from MariaDB.

mytile/ha_mytile.cc
DimitrisStaratzis: String attributes are supported now!

mytile/ha_mytile.cc
```diff
@@ -2527,7 +2756,38 @@ void tile::mytile::alloc_buffers(uint64_t memory_budget) {
...
       buff->offset_buffer = offset_buffer;
       buff->type = attr.type();
+    } else { // attribute with aggregate
+
+      tiledb::Attribute attr = this->array_schema->attribute(field_name);
+      uint8_t *validity_buffer = nullptr;
+      uint64_t *offset_buffer = nullptr;
+
+      uint8_t nullable = 0;
+      tiledb_attribute_get_nullable(ctx.ptr().get(), attr.ptr().get(),
```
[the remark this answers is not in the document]
DimitrisStaratzis: Yes there is, I have implemented it this way to keep the same style as the other cases in this method. I will switch to the C++ AP in all of them, it will make the implementation cleaner.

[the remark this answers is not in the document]
DimitrisStaratzis: Done!

robertbindar (changes requested):
Thanks for the great work @DimitrisStaratzis, I'm very sure this wasn't easy.
I do believe though that the current implementation is so limiting that it would be very hard to use in practice. This limitation comes mostly from using the MariaDB structures in a way they weren't intended to be used for this, I proposed in a comment below what I think the correct approach should look like.

I think this new approach should help remove most limitations you listed below. But there are 2 we need to clarify:
1) Can you please explain why aggregates won't work for joins?
2) I will investigate a bit further because it's a bit unclear to me the issue with data type promotion for some aggregates and if we really need the setting to get everything as double.

robertbindar:
> Thanks for the great work @DimitrisStaratzis, I'm very sure this wasn't easy. I do believe though that the current implementation is so limiting that it would be very hard to use in practice. This limitation comes mostly from using the MariaDB structures in a way they weren't intended to be used for this, I proposed in a comment below what I think the correct approach should look like.
> 
> I think this new approach should help remove most limitations you listed below. But there are 2 we need to clarify:
> 
> 1. Can you please explain why aggregates won't work for joins?
@DimitrisStaratzis ?
> 2. I will investigate a bit further because it's a bit unclear to me the issue with data type promotion for some aggregates and if we really need the setting to get everything as double.

After some more investigation, it seems to me that the issue isn't impossible to fix, MariaDB promotes the datatype of AVG results as well, please see `Item_sum_avg::create_tmp_field` in `server/sql/item_sum.cc`, seems like a very good point to start deep diving. It's either you need to store a sum and a count as string in the AVG item (and MariaDB will compute at the sql layer the average) or the current code is using the incorrect table_share (or incorrect field) to store the average result, use that function as a starting point, I'm sure you'll end up with an answer.
The problem with your approach in this PR isn't only the sysvar to special case the AVG function, it's the fact that if you'd need to average a huge number of integers, summing that huge number as FLOAT64 and dividing by the count in the end would give you a pretty different result than averaging over the same sequence of numbers as integers and this is not acceptable in my opinion for a database.

DimitrisStaratzis:
I am aware of the code repetition in the new ```init_scan()```. This will be resolved tomorrow.```std::cout```s will be removed as well of course.

mytile/ha_mytile.cc
```diff
@@ -131,6 +131,678 @@ tiledb::Context tile::build_context(tiledb::Config &cfg) {
...
+      this->ctx->handle_error(tiledb_array_get_non_empty_domain_from_index(
+          this->ctx->ptr().get(), aggr_array->ptr().get(), dim_idx,
+          nonEmptyDomain.get(), &empty_read));
+
+      nonEmptyDomains[dim_idx] = (std::move(nonEmptyDomain));
+    }
+  }
+
+  // if no ranges to add
+  // todo massive code repetition here, will create a method
```
DimitrisStaratzis, line 200: Will resolve this

DimitrisStaratzis: Done!

mytile/ha_mytile.cc
```diff
@@ -1053,6 +1397,47 @@ int tile::mytile::init_scan(THD *thd) {
...
+        DBUG_RETURN(std::nullopt);
+    }
+
+    // Get the current SELECT statement
+    SELECT_LEX *select_lex = thd->lex->current_select;
+
+    if (select_lex->with_sum_func) {
+      // Iterate through the item list to find TileDB compatible aggregates
+      Item *item;
+      List_iterator_fast<Item> it(select_lex->item_list);
```
robertbindar, line 1412: Please note that `select_lex` might not contain the aggregates in the item list, AFAIK that is supposed to be read from MariaDB `Query` passed at group by handler creation.

mytile/ha_mytile.cc
```diff
@@ -847,191 +1356,26 @@ int tile::mytile::init_scan(THD *thd) {
...
+                         this->subarray,
+                         &this->ctx,
+                         this->array.get());
 
     // If a query condition on an attribute was set, apply it
     if (this->query_condition != nullptr) {
       this->query->set_condition(*this->query_condition);
     }
 
-    this->subarray =
```
robertbindar, line 881: cheers for isolating this build subarray chunk into a standalone function.

mytile/ha_mytile.cc
```diff
@@ -847,191 +1356,26 @@ int tile::mytile::init_scan(THD *thd) {
...
-    for (uint64_t dim_idx = 0; dim_idx < this->ndim; dim_idx++) {
-      nonEmptyDomains.emplace_back(
-          std::unique_ptr<void, decltype(&std::free)>(nullptr, &std::free));
-    }
-
-    std::vector<std::pair<std::string, std::string>> nonEmptyDomainVars(
-        this->ndim);
-    for (uint64_t dim_idx = 0; dim_idx < this->ndim; dim_idx++) {
-      tiledb::Dimension dimension = domain.dimension(dim_idx);
+    this->subarray =
```
robertbindar, line 1359: make_unique?

mytile/ha_mytile.cc
```diff
@@ -177,6 +673,17 @@ int tile::mytile::external_lock(THD *thd, int lock_type) {
...
+tile::mytile_group_by_handler::mytile_group_by_handler(THD *thd_arg,
+     tiledb::Array *array,
+     std::shared_ptr<tiledb::Context>& context,
+     std::shared_ptr<tiledb::QueryCondition> &qc,
+     bool val_ranges,
+     bool val_in_ranges,
+     std::vector<std::vector<std::shared_ptr<tile::range>>> &ranges,
+     std::vector<std::vector<std::shared_ptr<tile::range>>> &in_ranges,
+     const std::string encryption_key,
+     const uint64_t open_at) : group_by_handler(thd_arg, mytile_hton), aggr_array(array), ctx(context), tiledb_qc(qc), valid_ranges(val_ranges), valid_in_ranges(val_in_ranges), pushdown_ranges(ranges), pushdown_in_ranges(in_ranges), encryption_key(encryption_key), open_at(open_at) {};
```
robertbindar, line 685: one intialization per line? it's very hard to follow now.

mytile/ha_mytile.cc
```diff
@@ -131,6 +131,501 @@ tiledb::Context tile::build_context(tiledb::Config &cfg) {
...
+  // Get the current SELECT statement
+  SELECT_LEX *select_lex = thd->lex->current_select;
+
+  if (select_lex->with_sum_func) {
+    // Iterate through the item list to find TileDB compatible aggregates
+    Item *item;
+    List_iterator_fast<Item> it(*query->select);
+    while ((item = it++)) {
+      Item_sum *isp = dynamic_cast<Item_sum *>(item);
+      if (isp && !field_is_aggregation_compatible(
```
robertbindar, line 601: I don't like suggesing names, but `field_is_aggregation_compatible` is pretty confusing, maybe try `aggregate_is_supported()`?

mytile/ha_mytile.cc
```diff
@@ -131,6 +131,501 @@ tiledb::Context tile::build_context(tiledb::Config &cfg) {
...
+                     isp->get_arg(0)->name.str,
+                     isp->sum_func(),
+                     aggr_array)) {
+          if (aggr_array != nullptr && aggr_array->is_open()){
```
robertbindar, line 605: just make `aggr_array` a `shared_ptr`, here and inside the group_by handler, it'll help you get rid of this resource release uggliness.

mytile/ha_mytile.cc
```diff
@@ -131,6 +131,501 @@ tiledb::Context tile::build_context(tiledb::Config &cfg) {
...
+        *ctx, uri, TILEDB_READ,
+        encryption_key.empty() ? TILEDB_NO_ENCRYPTION : TILEDB_AES_256_GCM,
+        encryption_key);
+#endif
+  }
+
+  // Get the current SELECT statement
+  SELECT_LEX *select_lex = thd->lex->current_select;
+
+  if (select_lex->with_sum_func) {
```
robertbindar, line 595: as I wrote below for `has_aggregates`, I don't think `thd->lex->current_select` is the intended way to get all aggregate items, it might be that we'll miss some cases and the only way to notice is performance degradation because MariaDB will do the aggregates anyway.
Also please refactor so that all the `while` loops looking for aggregates stay in a single function `has_aggregates` or something similar.

mytile/ha_mytile.cc
```diff
@@ -131,6 +131,501 @@ tiledb::Context tile::build_context(tiledb::Config &cfg) {
...
+}
+
+/**
+ * Checks if the given field in the given array is aggregation compatible
+ * @param field The input field
+ * @param aggregate The aggregation
+ * @param array_for_comp  The array
+ * @return
+ */
+static bool field_is_aggregation_compatible(const std::string field, const Item_sum::Sumfunctype aggregate, tiledb::Array* array_for_comp) {
```
robertbindar, line 495: name change suggested below.

mytile/ha_mytile.cc
```diff
@@ -131,6 +131,501 @@ tiledb::Context tile::build_context(tiledb::Config &cfg) {
...
+  // if it not an attribute, no TileDB aggregation can be applied
+
+  tiledb_datatype_t type;
+  tiledb::ArraySchema schema = array_for_comp->schema();
+  if (schema.has_attribute(field)){
+    auto attr = schema.attribute(field);
+    if (attr.cell_val_num() > 1 && !attr.variable_sized()) return false; // multi valued not supported
+    type = attr.type();
+  } else {
+    return false; // todo temporarily disable dim aggr pushdown shortcut-42339. Just remove after testing
```
robertbindar, line 505: I fixed this recently. Also aggregate on dimensions for dense was added but file as followup.

mytile/ha_mytile.cc
```diff
@@ -131,6 +131,501 @@ tiledb::Context tile::build_context(tiledb::Config &cfg) {
...
+        if (item_sum->sum_func() == Item_sum::MAX_FUNC) {
+          operation = std::make_unique<tiledb::ChannelOperation>(tiledb::QueryExperimental::create_unary_aggregate<tiledb::MaxOperator>(*aggr_query, column_with_aggregate));
+        } else {
+          operation = std::make_unique<tiledb::ChannelOperation>(tiledb::QueryExperimental::create_unary_aggregate<tiledb::MinOperator>(*aggr_query, column_with_aggregate));
+        }
+        std::string minmax_string = "minmax";
+        default_channel.apply_aggregate(minmax_string, *operation);
+        if (nullable) aggr_query->set_validity_buffer(minmax_string, validity);
+
+        switch (type) {
```
robertbindar, line 341: this is as nasty as it gets, please see if you can use `tiledb/type/apply_with_type.h` to get rid of this inner switch case over TileDB types or at the minimum put it in a separate function.

mytile/ha_mytile.cc
```diff
@@ -131,6 +131,501 @@ tiledb::Context tile::build_context(tiledb::Config &cfg) {
...
+
+        field->store(avg[0]);
+        break;
+      }
+      case Item_sum::MAX_FUNC:
+      case Item_sum::MIN_FUNC:
+      {
+        std::unique_ptr<tiledb::ChannelOperation> operation;
+        if (item_sum->sum_func() == Item_sum::MAX_FUNC) {
+          operation = std::make_unique<tiledb::ChannelOperation>(tiledb::QueryExperimental::create_unary_aggregate<tiledb::MaxOperator>(*aggr_query, column_with_aggregate));
```
robertbindar, line 333: break lines

robertbindar (changes requested):
This is great work @DimitrisStaratzis and it'll serve as a very good foundation for future aggregate support extensions and group bys, cheers!
Let me know if you need clarifications for any of the comments below.

mysql-test/mytile/t/aggregates.test
```diff
@@ -0,0 +1,149 @@
...
+DROP TABLE encr;
+
+##############################################################################
+## Test joins
+#joins
+--replace_result $MTR_SUITE_DIR MTR_SUITE_DIR
+--eval CREATE TABLE dense ENGINE=mytile uri='$MTR_SUITE_DIR/test_data/tiledb_arrays/1.6/quickstart_dense';
+select * from dense one JOIN dense two USING(a);
+select * from dense one JOIN dense two USING(a) where one.a > 10;
+select SUM(a) from dense one JOIN dense two USING(a) where one.a > 10;
```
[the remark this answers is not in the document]
DimitrisStaratzis: Yes, they do!

mytile/ha_mytile.cc
```diff
@@ -1986,6 +2371,11 @@ tile::mytile::cond_push_func(const Item_func *func_item,
   bool nullable = false;
   if (this->array_schema->has_attribute(column_field->field_name.str)) {
 
+    // Dense arrays do not support query conditions
```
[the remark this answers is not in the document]
DimitrisStaratzis: Query Conditions in Dense arrays are not allowed in mytile because the result buffers from TileDB, contain fill values for the filtered out values. So we don't allow them because a mariadb user is expecting to see only the values they requested. 

However, when we have an aggregate on a Dense array we do need to apply the query condition to the Query so that the aggregation produces the correct result.

mytile/ha_mytile.cc
```diff
@@ -1053,6 +1397,47 @@ int tile::mytile::init_scan(THD *thd) {
...
+      Item *item;
+      List_iterator_fast<Item> it(select_lex->item_list);
+      while ((item = it++)) {
+        Item_sum *isp = dynamic_cast<Item_sum *>(item);
+        if (isp) {
+          std::string column_with_aggregate = isp->get_arg(0)->name.str;
+          if (field == column_with_aggregate) {
+            switch (isp->sum_func()) {
+              case Item_sum::SUM_FUNC:
+                DBUG_RETURN(Item_sum::SUM_FUNC);
```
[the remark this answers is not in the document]
DimitrisStaratzis: It is. This function is only used to determine if we should push a QC to a dense array. COUNT in a dense array should not be pushed down because the QC does not fiter out results, but rather returns fill values. This causes the TIleDB count to count all values. I discussed this with @KiterLuc . This is the expecte dbehavior from core at the moment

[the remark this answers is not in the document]
DimitrisStaratzis: However, to make the code more clear, I have added the count case in this method and I check in ```cold_push_func()``` if it is a Dense & COUNT() case which we can;t cimpute at the moment

mytile/ha_mytile.cc
DimitrisStaratzis: Yes, this was a unique pointer before my PR, so I left it this way

mytile/ha_mytile.cc
DimitrisStaratzis: I know but this method is part of the mytile handler and it is only used when we create the query conditions. (They are created in the mytile handler). So I am currently unaware on how I can get access to the ```Query```

DimitrisStaratzis: The ```select_lex``` is taken from the thd so I don't see a reason as to why it might not include all aggregate items

mytile/ha_mytile.cc
DimitrisStaratzis: It crashes if I do that, I have no clue why...

mytile/ha_mytile.cc
DimitrisStaratzis: 1. see my response above
2. I don't see how I can combine all the different while loops, they all do very different things..

mytile/ha_mytile.cc
DimitrisStaratzis: refactored!

mytile/ha_mytile.cc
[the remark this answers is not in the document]
robertbindar: it makes sense, thanks.

mytile/ha_mytile.cc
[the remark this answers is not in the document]
robertbindar: and the QC disables computing the COUNT from frag meta if pushed down, I see now, thanks for the explanation.

mytile/ha_mytile.cc
robertbindar: can you get a stacktrace so maybe we catch fast why that's happening? I'd be nice to fix this before merging.

mytile/ha_mytile.cc
robertbindar: yeah, you're right, it's not as straightforward as I initially say, I'll mark as resolved, thanks.

robertbindar (approved):
Nice work! Let's try to figure out if we can fix the crash on `aggr_array` as shared_ptr, it'd be a shame to rush last minute and merge that thing, other than that, ready to merge!

mytile/ha_mytile.cc
DimitrisStaratzis:
```
0   mariadbd                            0x00000001067f3d3b my_print_stacktrace + 75
0   mariadbd                            0x00000001058338a7 handle_fatal_signal + 871
0   libsystem_platform.dylib            0x00007ff81741ec1d _sigtramp + 29
0   ???                                 0x0000600000000400 0x0 + 105553116267520
0   libsystem_c.dylib                   0x00007ff81733dca5 abort + 123
0   libc++abi.dylib                     0x00007ff8173ae082 abort_message + 241
0   libc++abi.dylib                     0x00007ff81739f225 _ZL28demangling_terminate_handlerv + 242
0   libobjc.A.dylib                     0x00007ff81709e02b _ZL15_objc_terminatev + 104
0   libc++abi.dylib                     0x00007ff8173ad4a5 _ZSt11__terminatePFvvE + 8
0   libc++abi.dylib                     0x00007ff8173ad456 _ZSt9terminatev + 54
0   ha_mytile.so                        0x0000000109c837a5 _ZN6tiledb5ArrayD1Ev + 21
0   ha_mytile.so                        0x0000000109c83785 _ZNSt3__112__destroy_atB6v15006IN6tiledb5ArrayELi0EEEvPT_ + 21
0   ha_mytile.so                        0x0000000109c83765 _ZNSt3__110destroy_atB6v15006IN6tiledb5ArrayELi0EEEvPT_ + 21
0   ha_mytile.so                        0x0000000109c83749 _ZNSt3__116allocator_traitsINS_9allocatorIN6tiledb5ArrayEEEE7destroyB6v15006IS3_vvEEvRS4_PT_ + 25
0   ha_mytile.so                        0x0000000109c827ce _ZNSt3__120__shared_ptr_emplaceIN6tiledb5ArrayENS_9allocatorIS2_EEE16__on_zero_sharedEv + 46
0   ha_mytile.so                        0x0000000109c30dd1 _ZNSt3__114__shared_count16__release_sharedB6v15006Ev + 49
0   ha_mytile.so                        0x0000000109c30d79 _ZNSt3__119__shared_weak_count16__release_sharedB6v15006Ev + 25
0   ha_mytile.so                        0x0000000109c78b1c _ZNSt3__110shared_ptrIN6tiledb5ArrayEED2B6v15006Ev + 44
0   ha_mytile.so                        0x0000000109c13f25 _ZNSt3__110shared_ptrIN6tiledb5ArrayEED1B6v15006Ev + 21
0   ha_mytile.so                        0x0000000109c5f9d5 _ZN4tile23mytile_group_by_handlerD2Ev + 117
0   ha_mytile.so                        0x0000000109c30765 _ZN4tile23mytile_group_by_handlerD1Ev + 21
0   ha_mytile.so                        0x0000000109c30789 _ZN4tile23mytile_group_by_handlerD0Ev + 25
0   mariadbd                            0x0000000105d5985c _ZN14Pushdown_queryD2Ev + 44
0   mariadbd                            0x0000000105d2d835 _ZN14Pushdown_queryD1Ev + 21
0   mariadbd                            0x0000000105d1907b _ZN4JOIN7cleanupEb + 1195
0   mariadbd                            0x0000000105d18985 _ZN4JOIN7destroyEv + 229
0   mariadbd                            0x0000000105df667e _ZN13st_select_lex7cleanupEv + 766
0   mariadbd                            0x0000000105ceb318 _Z12mysql_selectP3THDP10TABLE_LISTR4ListI4ItemEPS4_jP8st_orderS9_S7_S9_yP13select_resultP18st_select_lex_unitP13st_select_lex + 1688
0   mariadbd                            0x0000000105cea9e8 _Z13handle_selectP3THDP3LEXP13select_resulty + 552
0   mariadbd                            0x0000000105c7e663 _ZL21execute_sqlcom_selectP3THDP10TABLE_LIST + 2243
0   mariadbd                            0x0000000105c71128 _Z21mysql_execute_commandP3THDb + 6744
0   mariadbd                            0x0000000105c671c7 _Z11mysql_parseP3THDPcjP12Parser_state + 1015
0   mariadbd                            0x0000000105c63775 _Z16dispatch_command19enum_server_commandP3THDPcjb + 5365
0   mariadbd                            0x0000000105c68517 _Z10do_commandP3THDb + 3543
0   mariadbd                            0x0000000105eb4c76 _Z24do_handle_one_connectionP7CONNECTb + 406
0   mariadbd                            0x0000000105eb49ca handle_one_connection + 90
0   mariadbd                            0x000000010616af8f pfs_spawn_thread + 319
0   libsystem_pthread.dylib             0x00007ff8173f4259 _pthread_start + 125
0   libsystem_pthread.dylib             0x00007ff8173efc7b thread_start + 15
```

DimitrisStaratzis: I have absolutely no idea why this is happening..

DimitrisStaratzis: This is after just replacing the creation of the array from ```new``` to a ```shared_ptr```

DimitrisStaratzis: I think I am going to leave this as is for now, let me know if you have any objections. I will come back

PR #311, opened 2024-01-26
Update to TileDB-2.21.0 and add support for the TILEDB_GEOM_* types:

DimitrisStaratzis:


Shelnutt2 (approved):
Also @perrygeo what do we need to push the geometries down?

ihnorton (approved):

perrygeo:
Thanks! The geometry push down shouldn't be affected by the new types since we don't reference the geometry attribute directly; the spatial indexing happens on the integer dimensions and the geometry itself is treated as an opaque payload. That said, I owe you some tests to confirm, but it should be good to go.

PR #312, opened 2024-01-31
Simplify linking to tiledb.

teo-tsirpanis:
The `LINK_LIBRARIES` argument of `MYSQL_ADD_PLUGIN` apparently can accept the name of a CMake target. We directly pass `TileDB::tiledb_shared` instead of trying to extract the `IMPORTED_LOCATION` from the target, which will break with TileDB 2.20.

Validated by successfully building `Dockerfile-dev` on my machine.


Shelnutt2 (approved):

DimitrisStaratzis (approved):

DimitrisStaratzis:
Thanks @teo-tsirpanis

PR #313, opened 2024-02-06
add default case in ```build_metadata_value_string()```

DimitrisStaratzis:


DimitrisStaratzis:
Ubuntu nightly is green!

ihnorton (approved):
I'm ok with this, but it has the disadvantage that we don't get any alert if we forget to implement a type. It would be nice to find a different approach that would only do this for upcoming releases and still error out when we get to RC phase.

Shelnutt2:
> I'm ok with this, but it has the disadvantage that we don't get any alert if we forget to implement a type. It would be nice to find a different approach that would only do this for upcoming releases and still error out when we get to RC phase.

I believe we can set `-Wno-error=switch` for nightlies but leave `-Werror=switch` for the default/regular build?

PR #315, opened 2024-02-07
Revert "Simplify linking to tiledb."

teo-tsirpanis:
Reverts #312

Fixes [failures in the `Dockerfile-R` job](https://github.com/TileDB-Inc/TileDB-MariaDB/actions/runs/7738149768/job/21244630841).


teo-tsirpanis:
This will _not_ break when we update to 2.20, thanks to TileDB-Inc/TileDB#4671.

DimitrisStaratzis (approved):

DimitrisStaratzis (approved):

PR #316, opened 2024-02-07
Do not run Docker builds on forks

jdblischak:
The new GitHub Action workflow `docker` failed when run on my fork. I updated the workflow `stable` to only call `docker` when the repository owner is `TileDB-Inc`

Other notes:

* The `${{}}` aren't required in `if:`, eg [these examples](https://github.com/search?q=language%3Ayaml%20path%3A.github%2Fworkflows%20startsWith&type=code)
* The `if:` statement is now quite long. I can reformat it if desired, eg

```yaml
    if: |
      github.repository_owner == 'TileDB-Inc'
      && 
      (
        startsWith(github.ref, 'refs/tags/')
        ||
        github.ref_name == github.event.repository.default_branch
      )
```

### DIFF
--- a/.github/workflows/stable.yaml
+++ b/.github/workflows/stable.yaml
@@ -46,6 +46,6 @@ jobs:
         run: bash scripts/ci/build-tiledb-mariadb.sh
   deploy:
     needs: build
-    if: ${{ startsWith(github.ref, 'refs/tags/') || github.ref_name == github.event.repository.default_branch }}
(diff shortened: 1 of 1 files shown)


jdblischak:
This also disables the `docker` workflow in a PR from a fork, which I think makes sense since we don't want anyone to be able to build and upload Docker images just by submitting a PR (I suspect GitHub would probably prevent passing the credentials anyways, just causing it to fail)

Shelnutt2 (approved):

DimitrisStaratzis (approved):

PR #318, opened 2024-02-14
increase docker shortsha length

DimitrisStaratzis:
Will resolve https://github.com/TileDB-Inc/TileDB-MariaDB/actions/runs/7902170851/job/21575028022


ihnorton (approved):
(FYI - should be fixed in 2.20)

PR #332, opened 2024-04-08
Dstara/update to 2.22.0

DimitrisStaratzis:
This will be the first release of the aggregates pushdown. They will be enabled by default.


DimitrisStaratzis:
Discovered a small bug in aggregated, so delaying enabling it by default for the next release.

PR #346, opened 2024-06-05
Build TileDB in DockerFiles with debug info

DimitrisStaratzis:
Building TileDB from source and exporting it to the LD_LIBRARY_PATH and then using it in MariaDB does not work no matter what I try. I have tried everything, using the prefix, or even manually copy pasting the headers and the .so lib file to the directory the prebuilt artifacts where placed in the previous implementation.

This PR does the same thing in a different way. It forces MariaDB to build TileDB as a dependency. I have also modified the CmakeList.txt to build with ```-DCMAKE_BUILD_TYPE=RelWithDebInfo```. This has zero size-effects

Example of successful Dockerfile-min run: https://github.com/TileDB-Inc/TileDB-MariaDB/actions/runs/9513399605/job/26223431050#step:3:31780. The failure is because it failed to upload which was expected. 

[sc-49266]


docker/Dockerfile-min
```diff
@@ -73,10 +73,28 @@ RUN wget https://github.com/TileDB-Inc/TileDB-MariaDB/archive/${MYTILE_VERSION}.
...
+  -DTILEDB_AZURE=ON \
+  -DTILEDB_GCS=ON \
+  -DTILEDB_HDFS=ON \
+  -DTILEDB_SERIALIZATION=ON \
+  -DTILEDB_WEBP=ON \
+  -DTILEDB_TESTS=OFF \
+  -DVCPKG_TARGET_TRIPLET=x64-linux \
+  -DTILEDB_VCPKG_BASE_TRIPLET=x64-linux \
+  .. \
+&& make -j2 \
```
Shelnutt2, line 95: Can we do `make -j$(nproc)` everywhere?

docker/Dockerfile-min
```diff
@@ -65,18 +81,36 @@ RUN groupadd -r mysql && useradd -r -g mysql mysql \
...
+  -DTILEDB_GCS=ON \
+  -DTILEDB_HDFS=ON \
+  -DTILEDB_SERIALIZATION=ON \
+  -DTILEDB_WEBP=ON \
+  -DTILEDB_TESTS=OFF \
+  -DVCPKG_TARGET_TRIPLET=x64-linux \
+  -DTILEDB_VCPKG_BASE_TRIPLET=x64-linux \
+  .. \
+&& make -j$(nproc) \
+&& sudo make -C tiledb install \
```
teo-tsirpanis, line 112:
```suggestion
&& cmake \
  -DCMAKE_INSTALL_PREFIX=/usr/local/ \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DTILEDB_S3=ON \
  -DTILEDB_VERBOSE=ON \
  -DTILEDB_AZURE=ON \
  -DTILEDB_GCS=ON \
  -DTILEDB_HDFS=ON \
  -DTILEDB_SERIALIZATION=ON \
  -DTILEDB_WEBP=ON \
  -DTILEDB_TESTS=OFF \
  -DTILEDB_VCPKG_BASE_TRIPLET=x64-linux \
  .. \
&& sudo make -j$(nproc) tiledb install-tiledb \
```

* The vcpkg target triplet is not needed if specifying the base triplet.
* Updated the invocation of `make` to be future-proof when the superbuild gets removed, see TileDB-Inc/TileDB-Go#316

dudoslav (changes requested):
Can you post output of `make VERBOSE=1` or some proof that `libtiledb.so` is compiled with Debug symbols and is not stripped?

dudoslav:
For example, for TileDB-Vector-Search compiled with RelWithDebInfo build type we can see:
```
base) taco:vector_search dudoslav$ objdump -h _tiledbvspy.cpython-310-x86_64-linux-gnu.so | grep debug
 27 .debug_aranges 00018ff0  0000000000000000  0000000000000000  006ee25b  2**0
 28 .debug_info   0215f43e  0000000000000000  0000000000000000  0070724b  2**0
 29 .debug_abbrev 0000ba35  0000000000000000  0000000000000000  02866689  2**0
 30 .debug_line   0060d156  0000000000000000  0000000000000000  028720be  2**0
 31 .debug_str    02ef5fba  0000000000000000  0000000000000000  02e7f214  2**0
 32 .debug_line_str 000020c1  0000000000000000  0000000000000000  05d751ce  2**0
 33 .debug_loclists 00d6702c  0000000000000000  0000000000000000  05d7728f  2**0
 34 .debug_rnglists 0026651f  0000000000000000  0000000000000000  06ade2bb  2**0

and:

(base) taco:vector_search dudoslav$ file _tiledbvspy.cpython-310-x86_64-linux-gnu.so
_tiledbvspy.cpython-310-x86_64-linux-gnu.so: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=f0582f2ec9c3b5652afa951a2326eaa405505df8, with debug_info, not stripped
```

dudoslav:
I believe that we do not produce debug info because we use Dockerfiles that download MYTILE from releases and not from this branch. So after merge they should contain debug info, otherwise rework of Dockerfiles is needed. I would suggest to rework those Dockerfiles, maybe we could create a shortcut task for that.

dudoslav:
When using Dockerfile-dev that uses local changes I can see:
```
mysql@c5fa78c1a064:/usr/lib$ file libtiledb.so.2.24
libtiledb.so.2.24: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=98aa1eb6236d4ccad123623fabd3dceb0c1f7861, with debug_info, not stripped
```

So it seems like it should work, but keep in mind, that only after we change the release version in all dockerfiles.

dudoslav (approved):

DimitrisStaratzis:
Thank you @dudoslav for helping me with this!

PR #349, opened 2024-06-11
add support for Dense writes on Dense arrays

DimitrisStaratzis:
This PR add support for Dense writes in Dense arrays. This was not possible previously. To achieve this we needed to create a subarray with the given dim values. (https://docs.tiledb.com/main/background/internal-mechanics/writing#dense-writes)

I also added a test and a check to throw an error message when a user is trying to create a Dense array with a dim type other that ```MYSQL_BIGINT UNSIGNED```

[sc-49267]


mytile/mytile.cc
```diff
@@ -576,7 +576,16 @@ tile::create_field_attribute(tiledb::Context &ctx, Field *field,
...
 
 tiledb::Dimension tile::create_field_dimension(tiledb::Context &ctx,
-                                               Field *field) {
+                                               Field *field,
+                                               tiledb_array_type_t array_type) {
+
+  if (array_type == TILEDB_DENSE &&
+      ((field->type() != MYSQL_TYPE_LONGLONG) ||
+       !(dynamic_cast<Field_num *>(field))->unsigned_flag)) {
+    throw tiledb::TileDBError(std::string(
```
Shelnutt2, line 585: Is there a reason we limit this?

DimitrisStaratzis: In fact no. This error message has been changed and will be thrown only when creating a string dim in a dense array. Thanks 

https://docs.tiledb.com/main/how-to/arrays/creating-arrays/creating-dimensions

mytile/ha_mytile.cc
```diff
@@ -3249,6 +3251,32 @@ int tile::mytile::flush_write() {
...
+      if (buff->dimension && this->array_schema->array_type() ==
+                                 tiledb_array_type_t::TILEDB_DENSE) {
+        // In case of a dense write we need to create a subarray
+        // see
+        // https://docs.tiledb.com/main/background/internal-mechanics/writing#dense-writes
+
+        // Cast the dim buffer to uint64. This is the type for all dim buffers
+        // in dense arrays
+        uint64_t *uint64_buffer = static_cast<uint64_t *>(buff->buffer);
+        int num_elements = buff->buffer_size / 8;
```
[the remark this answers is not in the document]
Shelnutt2: I think you can use the buff->datatype to get the size of the type, then do pointer math to get the first and last value? We can then use the `void*` api for `add_range`?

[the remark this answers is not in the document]
DimitrisStaratzis: Yes! Will fix it first thing tomorrow

mysql-test/mytile/t/dense_writes.test
```diff
@@ -0,0 +1,27 @@
...
+CREATE TABLE test_dense (
+dim0 bigint UNSIGNED DIMENSION=1 lower_bound="0" upper_bound="100" tile_extent="10",
+attr1 varchar(255),
+attr2 int
+) engine=MyTile array_type='DENSE';
+
+INSERT INTO
+test_dense (dim0, attr1, attr2)
+VALUES
+(1, "cell 1", 100),
```
KiterLuc, line 13: Would it be possible to insert values that don’t have contiguous row ids? In this case, it might require multiple writes?

DimitrisStaratzis: In the current state it would require multiple writes yes. If we want to add it as a feature it would require to sort the insert buffers and construct the subarrays accordingly.

mytile/ha_mytile.cc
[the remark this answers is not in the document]
DimitrisStaratzis: Fixed!

mytile/ha_mytile.cc
```diff
@@ -3249,6 +3251,35 @@ int tile::mytile::flush_write() {
...
+      if (buff->dimension && this->array_schema->array_type() ==
+                                 tiledb_array_type_t::TILEDB_DENSE) {
+        // In case of a dense write we need to create a subarray
+        // see
+        // https://docs.tiledb.com/main/background/internal-mechanics/writing#dense-writes
+
+        // Get the first and last elements of the dim buffer
+        tiledb_datatype_t type = buff->type;
+        uint64_t type_size = tiledb_datatype_size(type);
+        int num_elements = buff->buffer_size / type_size;
```
Shelnutt2, line 3263: Question, for the write path, are these buffers sized exactly? Does this math work to the last value? For reads the buffers are fixed size but for writes do we just allocate however many records we have?

DimitrisStaratzis: Yes, they are populated here to the exact byte:

https://github.com/TileDB-Inc/TileDB-MariaDB/blob/0e9885fef573970fed7d8c52c1ee8f615a3af99f/mytile/ha_mytile.cc#L3123

which in turn calls: ```set_buffer_from_field()```

Shelnutt2 (approved):

mytile/ha_mytile.cc
DimitrisStaratzis: In fact to calculate the number of elements we use:

```mytile_buffer->buffer_size```

which is the "effective data buffer size". So we will never read out of memory

PR #355, opened 2024-07-05
fix Dockerfile download links

DimitrisStaratzis:
This PR will fix the main branch failures. The introduction of #346 is causing failures in the Dockerfile-R. The failures have to do with the fact the this is the only file that builds mariadb in "embedded" mode. This dockefile is not used anymore so it is removed by this PR. In addition to that, the download links for mariadb have been updated.

A successful run can be seen here https://github.com/DimitrisStaratzis/TileDB-MariaDB/actions/runs/9758760985/job/26934249792. The failures are because of the docker hub credentials. the images are built fine. 

[sc-50453]

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -22,6 +22,8 @@ jobs:
             repository: "tiledb/tiledb-mariadb-min"
           - dockerfile: "Dockerfile-server"
             repository: "tiledb/tiledb-mariadb-server"
+          - dockerfile: "Dockerfile-R"
+            repository: "tiledb/tiledb-mariadb-r"
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


dudoslav (approved):
If you were able to produce a container and tested out if everything work then feel free to merge

eddelbuettel:
> so it is removed by this PR

Per the diff here the Dockerfile-R is altered but not removed. It is however no longer _referenced_ from the YAML file.

DimitrisStaratzis:
> > so it is removed by this PR
> 
> Per the diff here the Dockerfile-R is altered but not removed. It is however no longer _referenced_ from the YAML file.

Yes, I decided to keep the file in case it is needed in the future. I also made sure it builds the same way it was being built before my changes. I am going to merge this. Thank you!

PR #367, opened 2024-08-19
Fix combining multiple operators

Shelnutt2:
This fixes a typo in combinding multiple operators to use `queryCondition` instead of `operatorCondition`.


Shelnutt2:
@DimitrisStaratzis would you mind looking at adding some unit tests for this based on the original query that failed but perhaps with a smaller subset of data?

DimitrisStaratzis:
> @DimitrisStaratzis would you mind looking at adding some unit tests for this based on the original query that failed but perhaps with a smaller subset of data?

Yes I am on it!

DimitrisStaratzis:
Fix is in #369. Closing this.

PR #373, opened 2024-09-03
enabling setting filters in dims and removing redundant code

DimitrisStaratzis:
~In the```discover_array()``` function in ```mytile_discovery.cc``` I see that we have code which does this:~

~https://github.com/TileDB-Inc/TileDB-MariaDB/blob/72ff60e97574995f2c3f21a96295b816802d6cce/mytile/mytile-discovery.cc#L420-L424~

~However, this method is only used when we want to read an existing array made outside of Mytile. In that case there is no need to specify the filters, the TileDB reader will read the data regardless. I have removed it and the test arrays are read fine even if they have filters.~

~This could speed up array discovery and eliminate a lot of redundant code.~

~In addition to the above~ I have added the ability to create dims with filters during the ```CREATE TABLE``` phase and use the default options in case the user doesn’t. [sc-54289]. The dimension filters are also shown when running ```SHOW CREATE TABLE```

### DIFF
--- a/mytile/mytile-discovery.cc
+++ b/mytile/mytile-discovery.cc
@@ -245,6 +245,27 @@ int tile::discover_array(THD *thd, TABLE_SHARE *ts, HA_CREATE_INFO *info) {
       table_options << " encryption_key=" << encryption_key;
     }
 
+    // Check for coordinate filters
+    tiledb::FilterList coordinate_filters = schema->coords_filter_list();
+    if (coordinate_filters.nfilters() > 0) {
+      table_options << " coordinate_filters='"
+                    << filter_list_to_str(coordinate_filters) << "'";
+    }
+
+    // Check for offset filters
+    tiledb::FilterList offset_filters = schema->offsets_filter_list();
+    if (offset_filters.nfilters() > 0) {
+      table_options << " offset_filters='" << filter_list_to_str(offset_filters)
+                    << "'";
+    }
+
+    // Check for validity filters
+    tiledb::FilterList validity_filters = schema->validity_filter_list();
+    if (validity_filters.nfilters() > 0) {
+      table_options << " validity_filters='"
+                    << filter_list_to_str(validity_filters) << "'";
+    }
+
     for (const auto &dim : schema->domain().dimensions()) {
       int mysql_type =
           TileDBTypeToMysqlType(dim.type(), false, dim.cell_val_num());
@@ -274,12 +295,17 @@ int tile::discover_array(THD *thd, TABLE_SHARE *ts, HA_CREATE_INFO *info) {
         sql_string << " dimension=1"
                    << " lower_bound='" << lower_domain << "' upper_bound='"
                    << upper_domain << "' tile_extent='"
-                   << dim.tile_extent_to_str() << "'"
-                   << ",";
+                   << dim.tile_extent_to_str() << "'";
       } else {
-        sql_string << " dimension=1"
-                   << ",";
+        sql_string << " dimension=1";
       }
+
+      // Check for filters
+      tiledb::FilterList filters = dim.filter_list();
+      if (filters.nfilters() > 0) {
+        sql_string << " filters='" << filter_list_to_str(filters) << "'";
+      }
+      sql_string << ",";
     }
 
 #if TILEDB_VERSION_MAJOR >= 2 && TILEDB_VERSION_MINOR >= 15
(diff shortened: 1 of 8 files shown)


mytile/mytile-discovery.cc
```diff
@@ -245,27 +245,6 @@ int tile::discover_array(THD *thd, TABLE_SHARE *ts, HA_CREATE_INFO *info) {
       table_options << " encryption_key=" << encryption_key;
     }
 
-    // Check for coordinate filters
-    tiledb::FilterList coordinate_filters = schema->coords_filter_list();
-    if (coordinate_filters.nfilters() > 0) {
```
robertbindar, line 250: I had to check in the code cause I didn't understand why this is there and how it fits with MariaDB storage engine architecture. This `coordinate_filters` attribute we set on `CREATE TABLE` has no active purpose in our codebase, but from what I undertstand it'll stand as the only way to see, from SQL, that this array has filters set on dimensions, i.e. via `SHOW CREATE TABLE` or `INFORMATION_SCHEMA.TABLES`.

mytile/mytile.cc
```diff
@@ -1398,91 +1398,17 @@ tiledb::FilterList tile::parse_filter_list(tiledb::Context &ctx,
...
       default: {
-        auto value = parse_value<int32_t>(f[1]);
+        auto value = f.size() > 1 ? parse_value<int32_t>(f[1]) : -1;
         filter.set_option(TILEDB_COMPRESSION_LEVEL, value);
         break;
       }
-      }
     }
+
     filter_list.add_filter(filter);
```
robertbindar, line 1407: what happens if `if.size() == 1`, does `filter` have a default value?

mytile/mytile.cc
```diff
@@ -1321,22 +1320,22 @@ tiledb::FilterList tile::parse_filter_list(tiledb::Context &ctx,
                       ME_ERROR_LOG | ME_FATAL, f[0].c_str());
     }
 
-    tiledb::Filter filter(ctx, filter_type);
-    if (f.size() > 1) {
```
robertbindar, line 1325: I would just put before the `switch`: `if (f.size() == 1) { continue; }`, we know `f` is at least size 1 before the `switch`, also we get rid of all the ternary operators from each case.

robertbindar (approved):
Functionality for adding filters on dimensions, OK.
Removing code for `coordinate_filter=`, I think that's a mistake.
Also please triple check, there is this behavior which we publicly documented: https://docs.tiledb.com/main/how-to/arrays/creating-arrays/setting-filters#setting-a-filter-list-for-all-dimensions
try to add a test to validate that per-dim filter overrides coordinates_filter= . @Shelnutt2 please chime in if that's something we don't support anymore.

mytile/mytile-discovery.cc
DimitrisStaratzis: Yes indeed the code I removed has no purpose in the code other than the one you mentioned: To be able to see if we have filters in dimensions/attributes. I did not think of that.

mytile/mytile.cc
DimitrisStaratzis: Yeah if ```size == 1``` then this condition is not true ```f.size() > 1```  so value will get the deafult value which is ```-1 ```

```auto value = f.size() > 1 ? parse_value<int32_t>(f[1]) : -1;```

File: mytile/mytile.cc (hunk 3 above)
Comment (DimitrisStaratzis) [reply mytile/mytile.cc:1325]: If we do that we will not be able to set a default value. This was a request from Stavros. To be able to do
```SQL
CREATE TABLE my_table (
    dim0 int tile_extent="10" filters="GZIP",
    dim1 int tile_extent="10",
    attr0 varchar(255),
    attr1 float,
    primary key(dim0, dim1) 
) engine=MyTile uri='{sparse_table_uri}'
```

and use the deafult value for GZIP, similar to python

Shelnutt2:
> Functionality for adding filters on dimensions, OK. Removing code for `coordinate_filter=`, I think that's a mistake. Also please triple check, there is this behavior which we publicly documented: https://docs.tiledb.com/main/how-to/arrays/creating-arrays/setting-filters#setting-a-filter-list-for-all-dimensions try to add a test to validate that per-dim filter overrides coordinates_filter= . @Shelnutt2 please chime in if that's something we don't support anymore.


You are correct @robertbindar, `coordinate_filter` is still supported and we should keep it here in MariaDB to avoid breaking change.

mytile/mytile-discovery.cc
Shelnutt2: > To be able to see if we have filters in dimensions/attributes. I did not think of that.

If this is used for the output of `SHOW CREATE TABLE` please keep it.

DimitrisStaratzis: Yeah I am reverting the change

mytile/mytile.cc
DimitrisStaratzis: Done. When not setting a value TileDB core uses the default values

mytile/mytile.cc
DimitrisStaratzis: See my comment below.

robertbindar (approved):
Thanks Dimitris!

robertbindar (approved):
Thanks Dimitris!

robertbindar (approved):
Thanks Dimitris!

PR #374, opened 2024-09-11
Fix leak of `mytile_group_by_handler::aggr_array`.

DimitrisStaratzis:
## Description
The group_by handler for aggregates opens the array we are querying as aggr_array to validate whether we can push down the aggregation to TileDB. If the pushdown isn't possible, we weren't closing the array, which caused the deletion issue

On top of that, Theo changed ```aggr_array``` to a smart pointer which is a nice bonus.

Thanks to @teo-tsirpanis for figuring out what the problem was!

## Validation of fix:

To validate this fix I used an already existing array ```theo```.

And created a test that runs a simple query on it:
``` sql
--replace_result $MTR_SUITE_DIR MTR_SUITE_DIR
--eval CREATE TABLE theo ENGINE=mytile uri='$MTR_SUITE_DIR/test_data/theo';

select * from theo; 

```

I inserted the following line in the close() method of the mytile handler:

```cpp
tiledb::Array::delete_array(this->ctx, this->uri);
```

The expected behavior is that the array should be deleted after the query. 

By running the test without the fix in this PR I get the following result along with an exception which means the the deletion was not successfull:

Screenshot 2024-09-11 at 2 17 04 PM

After the fix I get the expected result :

Screenshot 2024-09-11 at 2 18 12 PM

[sc-52768]

### DIFF
--- a/mytile/ha_mytile.cc
+++ b/mytile/ha_mytile.cc
@@ -149,7 +149,7 @@ int tile::mytile_group_by_handler::end_scan() {
 
   this->pushdown_ranges.clear();
   this->pushdown_in_ranges.clear();
-  delete this->aggr_array;
+  this->aggr_array.reset();
   DBUG_RETURN(0);
 }
 
@@ -170,7 +170,7 @@ int tile::mytile_group_by_handler::init_scan() {
     tile::build_subarray(thd, this->valid_ranges, this->valid_in_ranges,
                          empty_read, domain, this->pushdown_ranges,
                          this->pushdown_in_ranges, this->tiledb_sub,
-                         this->ctx.get(), this->aggr_array);
+                         this->ctx.get(), this->aggr_array.get());
   } catch (const tiledb::TileDBError &e) {
     // Log errors
     my_printf_error(ER_UNKNOWN_ERROR, "[init_scan] error for table %s : %s",
@@ -641,7 +641,7 @@ static group_by_handler *mytile_create_group_by_handler(THD *thd,
 
   // open array here before init scan because we need to check if the aggregate
   // requested can be processed by TileDB.
-  tiledb::Array *aggr_array;
+  std::unique_ptr<tiledb::Array> aggr_array;
   std::shared_ptr<tiledb::Config> cfg;
   std::shared_ptr<tiledb::Context> ctx;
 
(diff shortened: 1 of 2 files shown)


mytile/ha_mytile.cc
```diff
@@ -708,9 +708,14 @@ static group_by_handler *mytile_create_group_by_handler(THD *thd,
...
-        thd, aggr_array, ctx, qc, valid_ranges, valid_in_ranges, ranges,
-        in_ranges, encryption_key, open_at);
+        thd, std::move(aggr_array), ctx, qc, valid_ranges, valid_in_ranges,
+        ranges, in_ranges, encryption_key, open_at);
     return handler;
+  } else {
+    if (aggr_array != nullptr && aggr_array->is_open()) {
+      aggr_array->close();
+      aggr_array.reset();
+    }
```
teo-tsirpanis, line 718: This doesn't seem to be needed, the array will be closed and freed when the unique pointer gets out of scope.

DimitrisStaratzis: It seems that the aggr_array does not go out of scope and the array is never closed. All queries crash without this if.

mytile/ha_mytile.cc
```diff
@@ -650,32 +650,32 @@ static group_by_handler *mytile_create_group_by_handler(THD *thd,
...
             encryption_key.empty() ? TILEDB_NO_ENCRYPTION : TILEDB_AES_256_GCM,
             encryption_key.c_str()));
 #else
-    aggr_array = new tiledb::Array(*ctx, uri, TILEDB_READ,
-                                   encryption_key.empty() ? TILEDB_NO_ENCRYPTION
-                                                          : TILEDB_AES_256_GCM,
-                                   encryption_key, open_at);
+    aggr_array = std::make_unique<tiledb::Array>(
+        *ctx, uri, TILEDB_READ,
+        encryption_key.empty() ? TILEDB_NO_ENCRYPTION : TILEDB_AES_256_GCM,
```
KiterLuc, line 662: This could be moved outside of the if elses

DimitrisStaratzis: This is a constructor call with the "open_at" parameter. Not sure what you mean here. The other call (line 675) has a different set of parameters

KiterLuc: What I meant is the following code repeated 4 times:
`encryption_key.empty() ? TILEDB_NO_ENCRYPTION : TILEDB_AES_256_GCM`

Let's just extract in a variable before the if elses. It will make the make_unique calls more succinct and easier to read.

DimitrisStaratzis: Ah ok, yes done!

PR #379, opened 2024-09-18
Revert "Revert "Fix leak of `mytile_group_by_handler::aggr_array`. (#374)""

teo-tsirpanis:
Reverts TileDB-Inc/TileDB-MariaDB#376

It was determined that #374 is not the cause of the subsequently observed failures.


docker/Dockerfile
```diff
@@ -55,7 +55,7 @@ WORKDIR /tmp
 
 ENV MARIADB_VERSION="mariadb-11.0.2"
 
-ARG MYTILE_VERSION="0.34.1"
+ARG MYTILE_VERSION="0.34.0"
```
DimitrisStaratzis, line 58: The mytile version is downgraded here because we also included the bump in that PR. Not a big deal since we will merge the 0.34.2 bump after this but let's remove these changes

DimitrisStaratzis (changes requested):
The mytile version is downgraded here because we also included the bump in that PR. Not a big deal since we will merge the 0.34.2 bump after this but let's remove these changes

docker/Dockerfile
teo-tsirpanis: Done.

Shelnutt2 (approved):

PR #387, opened 2024-12-21
Fix broken version string.

NullHypothesis:
While updating REST to use our latest MariaDB release, I noticed two problems:

1. The version string in the Dockerfile had a typo, which broke the deploy job. This is fixed in d8d584c.
2. Unfortunately, the bison 2.7 brew package expired a few days ago, which also broke the macOS job. I removed the 2.7 pin (which resulted in a major version update), which seems to have worked.


teo-tsirpanis (approved):
Thanks!

NullHypothesis:
For what it's worth, I'm not authorized to merge this PR. Somebody else will have to.

